PR #1, opened 2012-03-21
Review - Add deleteOrphans() utility to the File class to remove unused files

greezybacon:
Also add a cron job to perform the task on demand
Change the implementation for Ticket::deleteAttachments to call this method
after simply removing all entries from the %ticket_attachment table


greezybacon:
Corrected as requested

PR #16, opened 2012-03-26
Add CSV (Excel) export option to ticket search results

greezybacon:
Save a session variable when the database query is executed for the ticket search page. When the [export] button is clicked,  retrieve the database query and create a CSV file.

PS - Don't forget about multiple browser tabs per session, perhaps showing different search results

### DIFF
--- a/include/class.export.php
+++ b/include/class.export.php
@@ -1,0 +1,135 @@
+<?php
+/*************************************************************************
+    class.export.php
+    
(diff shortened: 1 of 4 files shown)


protich:
Timestamp is a wrong word - I meant date e.g tickets-20120326.csv

greezybacon:
Date/timestamp it is

PR #25, opened 2012-03-26
Track history of tickets over time for snazzy dashboard reports

greezybacon:
Add a %ticket_history table to track ticket states over time, correlated to
the user account changing the ticket. Then, tickets are tracked to the table
when (re)opened, closed, overdue, transferred and assigned.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -723,6 +723,7 @@ class Ticket{
 
         $sql.=' WHERE ticket_id='.db_input($this->getId());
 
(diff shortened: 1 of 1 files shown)


greezybacon:
Properly track 'reopened' state

greezybacon:
Do we need an 'opened' history entry when the ticket is created? _(It is duplicate data, considering %ticket::created will have the same data)_

_Might need that for tracking tickets created per timeframe_

PR #79, opened 2012-04-25
SetupWizard

protich:
**Installer**
Added checksum validation 
Fixed syntax error on ticket_event table - sql schema

**Upgrader**
Added debug and error logging 

**Misc**
Added a transitional patch - for devs 
Changed system logs default sort from date to primary key.


greezybacon:
I would vote to rename `abe9c0cb-bbb021fb.patch.sql` to `abe9c0cb-49478749.patch.sql`, and just change the signature in the patch and in `main.inc.php` to the new hash.

setup/inc/sql/osticket-v1.7-mysql.sql
```diff
@@ -615,7 +615,7 @@ CREATE TABLE `%TABLE_PREFIX%ticket_event` (
   `topic_id` int(11) unsigned NOT NULL,
   `state` enum('created','closed','reopened','assigned','transferred','overdue') NOT NULL,
   `staff` varchar(255) NOT NULL default 'SYSTEM',
-  `annulled` tinyint(1) unsigned NOT NULL defalt '0',
+  `annulled` tinyint(1) unsigned NOT NULL default '0',
```
greezybacon, line 618: My bad. Sorry, dude.

protich:
Renaming the path is fine but would require resetting any current installations directly in the DB - otherwise the upgrader won't know what to do. 

I'm more worried about testers who may be working off develop branch.

setup/inc/sql/osticket-v1.7-mysql.sql
protich, line 618: Don't sweat it bro! I should have done a test installation during review. Adding it to the checklist.

PR #81, opened 2012-04-26
Flexible filter settings

greezybacon:
**This code is completely untested**

I just want to start the conversation. This replaces the columns in the email_filter table with rows in the email_filter_setting table. It will pave the way for more flexible and creative settings like canned-response to apply to the new ticket.


protich:
If it's untested - I assume it won't cut dpr3 release (which I'll be releasing in the morning)? I'm asking because of the version you embedded on the patch.

PS. I like the idea. I think we had discussed about doing the same thing with config table.

greezybacon:
Yeah, sorry. Please don't release it with dpr3.

protich:
Closing it for now - we'll revisit the idea post 1.7 release.

PR #103, opened 2012-05-08
Feature/ajax urls revisited

protich:
Change ajax calls to use tickets/ instead of ticket/ to match api calls


greezybacon:
I still vote for `GET /ajax.php/tickets` for a search operation rather than `GET /ajax.php/tickets/search`

protich:
Totally forgot about your suggestion!

Thanks,

Issue #98, opened 2012-05-08
Cannot print Tickets


leovipin1:
Hi,

When I open a ticket and to print the ticket when i click on the print ticket button, it doesn't print the ticket.

kindly look into the same as well.

protich:
Still in development... working on PDF export.

greezybacon:
leovipin1, does the system print to your liking with the latest development snapshot?

leovipin1:
The ticket is being exported in pdf and now able to print

PR #114, opened 2012-05-27
PDF Export

protich:
Add ability to export ticket view as PDF (ticket print view).
- Issue #98


greezybacon:
Looks great. Consider pulling extra commits from my repo in feature/ticket-print

PR #99, opened 2012-06-18
Shows all closed tickets

greezybacon:
Hi,

When all the open tickets are closed then it shows all the closed tickets.

It must show "query returned 0 results" instead of closed tickets as it creates confusion a lot of time. This was a drawback in the 1.6 version. Kindly rectify this in the 1.7 version release.


greezybacon:
I disagree with the 'query returned ...' statement. I would recommend something like _There are no open tickets! Way to go! (Get yourself some coffee)_

PR #93, opened 2012-06-18
new canned response not working

greezybacon:
When trying to add a new canned response this is comes up with the fallowing error:

Unable to add canned response. Correct error(s) below and try again.

Status is Active
Tile: test
Canned Response: testing123
Nothing is attachment and internal notes.


greezybacon:
Thanks for reporting this!

PR #134, opened 2012-06-23
Ticket assign to team

greezybacon:
osTicket (v1.7-DPR4)
PHP: 5.4.0-3

If you assign a ticket to a team, you can not view this ticket

You will get an Fatal error:
Fatal error: Call to a member function getId() on a non-object in /include/staff/ticket-view.inc.php on line 23

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -20,7 +20,10 @@ $lock  = $ticket->getLock();  //Ticket lock obj
 $id    = $ticket->getId();    //Ticket ID.
 
 //Useful warnings and errors the user might want to know!
-if($ticket->isAssigned() && $staff->getId()!=$thisstaff->getId())
+if($ticket->isAssigned() && (
+            ($staff && $staff->getId()!=$thisstaff->getId())
+         || ($team && !$team->hasMember($thisstaff))
+        ))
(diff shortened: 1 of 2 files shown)


kurim:
Thank you greezybacon

works perfectly

^fh

greezybacon:
We should merge the attched code to the develop branch before closing

include/staff/ticket-view.inc.php
```diff
@@ -20,7 +20,10 @@
 $id    = $ticket->getId();    //Ticket ID.
 
 //Useful warnings and errors the user might want to know!
-if($ticket->isAssigned() && $staff->getId()!=$thisstaff->getId())
+if($ticket->isAssigned() && (
+            ($staff && $staff->getId()!=$thisstaff->getId())
+         || ($team && !$team->hasMember($thisstaff))
```
protich, line 25: The warning was meant to be displayed ONLY if a ticket is assigned to someone else..  I guess it doesn't hurt but to me warning about an active lock on a ticket (next else) is more important than the team. 

May be we actually need to move the lock warning to the top.

greezybacon, line 25: The warning still works the way you describe. It checks to see if a staff owns the ticket and it is a staff other than the current staff, otherwise, if a team owns the ticket and the current staff is not a member of the team. If either is true, then the warning is shown.

I agree that the lock warning is likely more important than the ownership warning. Should we show both?

protich, line 25: I'll merge it as it is - we'll move the lock warning higher when I revisit system warnings.

PR #111, opened 2012-06-24
cron.php - 500 Internal Server Error

greezybacon:
Hi there,

I am not able to execute cron.php, page does not load, cancels with 500 - Internal Server error

Thanks in advance!

Borstix

### DIFF
--- a/api/.htaccess
+++ b/api/.htaccess
@@ -1,4 +1,3 @@
-Options +FollowSymLinks
 RewriteEngine On
 
 RewriteBase /api/


greezybacon:
I feel like a total noob here, but how are you guys executing the `cron.php` script? In practice and in testing? -- I mean, how are you testing the `cron.php` script to get the HTTP/500 error, and how do you plan on triggering `cron.php` in production?

The basis of my question is that the script works fine for me, but sends an HTTP/403 with an extended error code 77 (EX_NOPERM), because I'm supposed to send the API key in the User-Agent HTTP header (which is kinda strange). So I was wondering how folks use the script in production environments (crontab or HTTP request).

PR #145, opened 2012-07-02
2 letter username support

Bastian-Kuhn:


Bastian-Kuhn:
next try with github... sorry

PR #146, opened 2012-07-07
Upgrader revisited

protich:
- Upgrader is now part of admin panel.
- Support upgrade from v1.6 RC1-5 to v1.7 X


greezybacon:
I made a couple last-minute commits to my branch. Please consider adding in those to the list.

protich:
Your commits merged to the PR.

include/class.pdf.php
```diff
@@ -17,7 +17,7 @@ class.pdf.php
 define('THIS_DIR', str_replace('\\\\', '/', realpath(dirname(__FILE__))) . '/'); //Include path..
 define('FPDF_DIR', THIS_DIR . 'fpdf/');
 define('FPDF_FONTPATH', FPDF_DIR . 'font/'); //fonts directory.
-session_cache_limiter('private');
+//session_cache_limiter('private');
```
greezybacon: I would prefer to see this line simply removed. Since we use Git, we have a record of it and could get it back. No need tempting ourselves to uncomment this line accidentally in the future

greezybacon:
This looks good. No lint errors, and an upgrade from this branch still goes without a hitch. Please pull the commit I added for the files-to-be-removed list.

include/staff/header.inc.php
```diff
@@ -2,6 +2,8 @@
 <html>
 <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8">
+    <meta http-equiv="cache-control" content="no-cache" />
+    <meta http-equiv="pragma" content="no-cache" />
```
greezybacon, line 6: These lines look like guesses toward the cache issue, which was really something else. However, they seem correct, permit-able and possibly useful. So it's fine with me to leave them in

protich, line 6: It **wasn't** guess work. It's a **fallback** for when caching is enabled AND _safe_mode_ is turned on  (which renders init_set calls useless). It doesn't work in all browsers  e.g safari as long as http headers sent back enable caching.

PR #148, opened 2012-07-09
Ticket Print Options

protich:
Staff can now print a ticket thread with or without internal notes + select paper size on the fly


greezybacon:
Looks like there's a UTF-8 issue with the PDF printing -- such as international names. It's not related to these commits. If I can't find it in a few minutes, I'll post a separate issue

protich:
Good point - I'll vote on a separate issue.

greezybacon:
Note to self | selves, the print form will need CSRF protection

greezybacon:
I made a couple of style tweaks. Mainly to separate the [Print] and [Cancel] buttons, and to make the dialog feel more like a common software dialog. Lemme know what you think. Otherwise, it looks ready to pull.

protich:
Changes merged.

PR #153, opened 2012-07-10
bug with no-break space

Bastian-Kuhn:
In some cases the no-break space crashes the mysql insert when creating a ticket

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -114,6 +114,8 @@ class Format {
     }
 
     function striptags($var) {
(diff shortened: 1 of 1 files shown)


protich:
While the idea is sound - you implemented it in the wrong spot!

Format::striptags should remain generic as it is. The fix should go to class.mailfetch.php => getBody function.

Thanks,

Bastian-Kuhn:
Ok, you are right

PR #157, opened 2012-07-12
Fix eMail attachments

Bastian-Kuhn:
Fixed the saving of Attachments from eMails.


greezybacon:
Thanks for catching this

PR #158, opened 2012-07-12
Fix removed no-break spaces from emails

Bastian-Kuhn:
In a few cases the html nbsp sign crashes the inserting of mail body into the database.


Bastian-Kuhn:
That seems to work. Thank you!

PR #159, opened 2012-07-13
Automaticly showing inline Images in Post Body

Bastian-Kuhn:
Showing of inline images in body text from ticket posts.

For us a very useful future, cause many customers send us inline attachments with screenshots from the problems they have.


scp/css/scp.css
```diff
@@ -1318,6 +1318,10 @@ time {
     border:1px solid #a00;
 }
 
+.ineline_image {
```
greezybacon: Did you misspell this no purpose?

greezybacon:
I really like this idea. I would prefer to see
- attachments inlined by the MIME content-type instead of the extension
- the list of inlined attachment types configurable (in `ost-config.php` or via the web interface)
- the `Ticket::getAttachmentUrls()` method merged with `Ticket::getAttachmentsLinks()` so we don't have yet another place where attachment URL hashes are generated

Bastian-Kuhn:
Hello,

i'm currently for a consulting in malysia,
i will fix all at the latest,  when i'm back in germany.

Bastian-Kuhn:
Hello, 

i just fixed the typos and merged the getAttachmentUrls wird the getAttachmentsLinks function.

For attachments inlined by the MIME content-type instead of the extension: I think that the effort here is to high in relation to the advantage, because we only deal with images an no other filetypes. (If you really want it this way, i change it of course)

the list of inlined attachment types configurable (in ost-config.php or via the web interface:
Yes, but i think an simple on or off are enough. We currently only have images here.
For a config option via web interface, it needs an update in the database scheme. How can i apply such an update?

n0rb:
why has this never been merged?

greezybacon:
This is a major feature of osTicket/osTicket-1.8. It creates several cascading issues, so we decided to leave it out of 1.7. Users who wish not to use the HTML feature can remain on osTicket 1.7

PR #150, opened 2012-07-13
Canned Autoresponse

greezybacon:
I merged the previous PR by mistake - I meant to add the following comments;
1. (Automatically respond with this canned attachment) - > attachment => response?
2. Make sure we restrict deletion of in-use canned response in knowledge base (check filters)

### DIFF
--- a/include/staff/cannedreplies.inc.php
+++ b/include/staff/cannedreplies.inc.php
@@ -2,12 +2,10 @@
 if(!defined('OSTSCPINC') || !$thisstaff) die('Access Denied');
 
 $qstr='';
-$sql='SELECT canned.*, count(attach.file_id) as files, dept.dept_name as department, '.
-     ' (filter.id is not null) as in_use '.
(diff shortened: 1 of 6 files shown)


include/staff/cannedreplies.inc.php
```diff
@@ -81,9 +83,12 @@
                 ?>
             <tr id="<?php echo $row['canned_id']; ?>">
                 <td width=7px>
+                <?php if (!$row['in_use']) { ?>
```
protich: Does this means it can't be disabled - when in-use by a filter?

I'll make a case that it's OK to allow admin to DISABLE it as away to turn off a canned-auto-response used by multiple filters for example.

The proper place to protect against deletion is `$canned->delete()` function. Do a join on load and simply do `if($this->isInUse()) return false;`

protich:
Having the actual count of filters might be helpful - we can use it on canned response page as part of the warning.

PR #161, opened 2012-07-16
Feature/dept access

protich:
Group's department access moved to its own table.
Ability to add/remove access on department add/edit.


include/class.dept.php
```diff
@@ -47,7 +50,7 @@ function load($id=0) {
         $this->id=$this->ht['dept_id'];
         $this->email=$this->sla=$this->manager=null;
         $this->getEmail(); //Auto load email struct.
-        $this->members=array();
+        $this->members=$this->groups=array();
```
greezybacon, line 53: Won't this clobber `$this->members` and `$this->groups`?

protich, line 53: You call it clobber - I can it init.

greezybacon:
What about the leftovers in `setup/inc/sql`. When do you want to clean those up?

include/upgrader/sql/435c62c3-2e7531a2.patch.sql
```diff
@@ -0,0 +1,31 @@
...
+        `updated`=IF(TO_DAYS(`updated`), `updated`, NOW());
+
+-- Fix groups dates... 
+UPDATE `%TABLE_PREFIX%groups` 
+    SET `created`=IF(TO_DAYS(`created`), `created`, IF(TO_DAYS(`updated`), `updated`, NOW())),
+        `updated`=IF(TO_DAYS(`updated`), `updated`, NOW());
+
+-- Finished with patch
+UPDATE `%TABLE_PREFIX%config`
+    SET `schema_signature`='6007d45b580c6ac0206514dbed0f28a6';
```
greezybacon: Fail. Should be `2e7531a201b5b8650dcd43681a832ebd`. This has an interesting side effect, though: it spins the upgrader into some strange loop. It starts the upgrade which dumps the `schema_signature` at some point from which there is no patch. The upgrader should crash and say the upgrade is not possible; however, instead viscously and continuously calls the upgrader, which responds with:

```
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Connection: close
Content-Encoding: gzip
Content-Type: text/html; charset=UTF-8
Date: Sun, 22 Jul 2012 03:41:17 GMT
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Pragma: no-cache
Server: nginx
Status: 200 OK
Transfer-Encoding: Identity

Upgrade osTicket to 1.7-DPR4
```

What do you think? Strange, eh?

protich: Not sure how I missed that one - I'll fix the upgrader issue too - it should be a hard fail.

protich:
Ready for another round of review.

greezybacon:
There are some more issues, but they can wait for 1.7.1

protich:
What issues?

greezybacon:
Couple typos and visual style tweaks

protich:
It might make sense to fix the typos now... just log an issue and I'll add it to my todo list.

include/staff/department.inc.php
```diff
@@ -191,13 +205,36 @@
         </tr>
         <tr>
             <th colspan="2">
+                <em><strong>Department Access</strong>: Check all groups allowed to access department.</em>
+            </th>
+        </tr>
+        <tr><td colspan=2><em>Primary department members and manager will always have access regarless of group selection or assignment.</em></td></tr>
```
greezybacon, line 211: `regar[d]less` is misspelled

include/staff/department.inc.php
```diff
@@ -191,13 +205,36 @@
...
+             .' LEFT JOIN '.STAFF_TABLE. ' staff USING(group_id) '
+             .' GROUP by grp.group_id '
+             .' ORDER BY group_name';
+         if(($res=db_query($sql)) && db_num_rows($res)){
+            while(list($id, $name, $members) = db_fetch_row($res)) {
+                if($members>0) 
+                    $members=sprintf('<a href="staff.php?a=filter&gid=%d">%d</a>', $id, $members);
+
+                $ck=($info['groups'] && in_array($id,$info['groups']))?'checked="checked"':'';
+                echo sprintf('<tr><td colspan=2>&nbsp;&nbsp;<label><input type="checkbox" name="groups[]" value="%d" %s>&nbsp;%s</label> (%s)</td></tr>',
```
greezybacon, line 224: I would prefer to see more than one column of groups. This will make the page shorter and make the configuration appear less daunting should there be several groups available to pick from.

include/staff/department.inc.php
```diff
@@ -191,13 +205,36 @@
         </tr>
         <tr>
             <th colspan="2">
+                <em><strong>Department Access</strong>: Check all groups allowed to access department.</em>
```
greezybacon, line 208: I don't feel that this statement is clear / intuitive. We've talked several times about this feature, and when i read this, I'm not really sure what exactly this means. _Truthfully, I think we need to start some detailed documentation on how to properly configure and utilize osTicket_

include/staff/department.inc.php
```diff
@@ -141,7 +144,18 @@
...
-        <?php } ?>
+        <?php 
+        } ?>
+
+        <tr>
+            <td width="180">
+                Group Membership:
+            </td>
+            <td>
+                <input type="checkbox" name="group_membership" value="0" <?php echo $info['group_membership']?'checked="checked"':''; ?> >
```
greezybacon, line 155: This checkbox should visually hide or disable the group selection since it disables the feature. _Unless I just totally don't get it_

include/staff/department.inc.php
greezybacon, line 211: `manager[s]` should probably be pluralized -- it will flow better with plural `department members`

include/staff/department.inc.php
protich, line 224: I agree - I plan on reworking UI a little down the road.... That said, I doubt anyone will have more than a dozen groups and in addition it's a page you deal with it only once.

include/staff/department.inc.php
protich, line 208: Feel free to suggest a better wording....  I totally agree on the docs - I already have a skeleton version.

include/staff/department.inc.php
protich, line 211: I'll fix the typo. 

My thinking was - it's in the context of dept. A department can only have one manager - managers will imply that you can have more than one manager.

include/staff/department.inc.php
protich, line 155: No - the checkbox basically extends department's membership to also include groups with access.  It doesn't change the group's access. All it means is - anyone with access to the department will get alerts and such. 
- See `getMembers` in class.dept.php

PR #173, opened 2012-07-27
Emailtest fixes and enhancements

soif:
Fixes
- Fields are now populated on the success page

Features:
- Emails can be "email@domain.com" or "Name &lt;email@domain.com&gt;"
- A new Incoming form is added. It allow to perform test email posted to OsTicket on the behalf on an external client. As a new OsTicket user, this form has proved to be very usefull to QUICKLY test if all is working fine.

Please note that I included js and css in the HTML: I know this is bad, but this allow you to test and accept the pull as-is.

If accepted I swear that I will quickly :
- move js code to the external js file (Just tell me if you're OK to use the jquery UI Tabs: in this case tell me what custom jquery UI widget you've currently included in the "jquery-ui-1.8.18.custom.min" file, so that I can rebuild it correctly, while adding the Tabs Widget)
- move the style to the scp.css stylesheet: Depending on your acceptation of my previous CSS pull request, i will add the style to the current CSS, or to my CSS. Current embeded style should work on both yours or my css.

BTW: I'm still spending more hours gitting than coding :-/ Don't hesitate to advise me on how you'd except a pull request, if i'm doing anything bad


soif:
I just added a fetch email accounts tab, that allow to force mail fetching, regardless of the last fetch.
(Usefull to help debugging faster)

protich:
I struggling to see the need and purpose of incoming email form; If you need to test incoming emails then simply email one of the system emails.

soif:
For a experienced user like you, there is perhaps no need. For new comers or developers like me, this is really useful :-/

Yes I can simply send a email, but when i tested the system i had to send dozens (if not hundred) of emails to watch if alert are sent correctly, if Ticket are correctly assigned etc......

Doing this with an email client, force you to fill each time the To, Subject and Message fields, and you can't do that on the behalf on another From email address.
When sending dozen of emails this is really boring and time consuming.

My new form (because it populates the form fields after each submit) make it really easy and fast. 
You can send emails again and again, while also being able to fastly change one of the field (ie the Subject or To)....

protich:
I don't think sending and email requires special experience - in fact it provides a real world test given how different email clients formats texts.

While I see your point, I don't think it's enough of a justification to bloat osTicket with unnecessary & confusing features. osTicket users will most likely assume that is a way to open tickets on behalf of customers. 

I like the idea of testing on demand email fetching. I'll recommend creating an issue - to start the conversation and how to go about implementing it - we're leaning on going with bootstrap css for UI look and feel.

soif:
I understand your point, i will keep it only for private use.

Issue created:
https://github.com/osTicket/osTicket-1.7/issues/175

PR #179, opened 2012-07-31
Hotfix/fix faq on submit

soif:
fix template name typo, giving an ugly include error on faq submit


protich:
The fix is correct - `kb-category.inc.php` doesn't exist. The only issue with this PR is the fact that it was submitted as a hot fix.

greezybacon:
@soif, can you verify that #220 fixes the issue?

soif:
confirmed, issue fixed

PR #180, opened 2012-07-31
Feature/Client auto login

soif:
Auto login the client when he clicks in the link provided in the email auto-response, instead of letting him in front of a confusing "Authorization Required" (auto-filled) form.


soif:
re submitted correctly to the develop branch

protich:
Auto login was not implemented by design - so the form can be bot & CSRF protected. It's not foolproof but helps with brute force.

PR #181, opened 2012-07-31
Feature/Client auto login

soif:
Auto login the client when he clicks in the link provided in the email auto-response, instead of letting him in front of a confusing "Authorization Required" (auto-filled) form.

### DIFF
--- a/include/client/login.inc.php
+++ b/include/client/login.inc.php
@@ -3,7 +3,20 @@ if(!defined('OSTCLIENTINC')) die('Kwaheri');
 
 $email=Format::input($_POST['lemail']?$_POST['lemail']:$_GET['e']);
 $ticketid=Format::input($_POST['lticket']?$_POST['lticket']:$_GET['t']);
+
+if($_GET['e'] && $_GET['t']){
 ?>
+	<script language="javascript">
+	$(document).ready(function(){
+		if($('#email').val() && $('#ticketno').val()){
+			$('#clientLogin').submit();
+		}
+	});
+	</script>
+<?php
+}
+?>
+
 <h1>Check Ticket Status</h1>
 <p>To view the status of a ticket, provide us with the login details below.</p>
 <form action="login.php" method="post" id="clientLogin">


protich:
See my comments on #180

soif:
IMO letting the client (after he clicked on a valid link) on an page that look like an error ("Authorization required" in red), is not really the most nice experience, we can offer him... ;-)

Would it seems more acceptable to you, if once the client is on this page, I use javascript to automatically submit the form, if $_GET are filled ?

soif:
Changed to Javascript auto submit. Is this now OK?

soif:
I'm not sure to fully understand what you mean in your latest comment (sorry english is not my native language). So, sorry if I missed your point:

I you look at the pull-request , in the diff, you'll see that currently, the only final added code, is a little javascript that auto submit the form, if $_GET vars are filled. Is this not OK for you?

About your commented code line, I had initially changed the clientLoginPage to the "require login.php", because, there is at least an issue I encountered while implementing my first php solution to auto-login: 

Because login.php is now called from a function, all global variables are not defined.
For example, $cfg and $ost (used in login.php) are not defined if login.php is included from a function, instead of beeing directly included. If you keep the require in the clientLoginPage as-is, IMHO, you should at least put a "global $cfg, $ost;" before the include, to prevent future bugs to happen. See what I mean?

protich:
My point was you can not simply change code to make your stuff work without considering why it was there in the first place. Your changes basically breaks ajax interface - if you need the variables THEN add them.
- Honestly I don't like the client side based auto login. Once deployed you can not disable it (it's not optional). My thinking is to introduce an `%auth` variable which can also be used as part of CC/BCC implementation.

soif:
But what's wrong with auto-login the client, instead of having him to click the form button? 
This is also -not- optional at this time...

protich:
I've made my case here and on issue #180. You're welcome to disagree.

soif:
I really don't understand what is wrong with auto-login the client, now that my feature is bot & CSRF protected as you wished (issue #180). I thought that I had followed you wishes.

Please understand that,  on my side (and without any developer guidelines..), trying to understand what you want or not, is really not clear, if you don't explain me more.

greezybacon:
So what if osTicket added some auth token to the mix (e=<email>&t=<ticket>&a=<auth-token>), where the auth-token is something simple like `MD5(email + ticket# + SECRET_SALT)` so that the system can easily verify that it originated the email and the link. It's bot and brute-force proof and doesn't break CSRF protection.

We can discuss the `%auth` variable in more detail as well

protich:
@greezybacon - That is exactly what I was suggesting above -  `%auth`  **is**  auth-token! This allows admin to determine which department (based on email template - for example) can auto-login. This is especially important - once we add optional captcha to login page. `%auth` will help authenticate/verify the link  and by-pass the login page even IF captcha is enabled.

soif:
rebased on rc2 develop, now only javascript auto-login

protich:
@soif - Jared already implemented %auth based auto-login. We're both not in favor of javascript based solution for reasons stated above.

PR #185, opened 2012-08-01
Fix filters applied on web created tickets

soif:
If we create a TT from the web (no 'emailId'), email filters with an emailId defined, are still applied to the ticket. This is obviously not wanted.

This fix prevents this bug to happen.


protich:
@greezybacon - Looks good to me - checking the source makes sense but not necessary. Yes - I still plan on moving filters to settings tab and allowing for an option to limit to specific source (All, web, email (all), specific email address or API)

greezybacon:
Boo. Didn't notice the pull was for the master branch!

PR #186, opened 2012-08-01
Fix Quoted Reply not removed

soif:
Quoted part of an email response is now properly removed


greezybacon:
Thanks, soif. I appreciate you attention to detail

soif:
You're welcome :-)

Issue #190, opened 2012-08-01
Mime Types validation


protich:
Improve attachment validation to include mime types

soif:
like this?
https://github.com/osTicket/osTicket-1.7/pull/196

protich:
No - meant validation on incoming attachments.

PR #196, opened 2012-08-02
Mime Types (Issue #190)

soif:
Set MimeType based on file extension, instead of trusting the web browser uploading the file.

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -107,8 +107,9 @@ class AttachmentFile {
 
     function display() {
        
-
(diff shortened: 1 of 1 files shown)


soif:
This could surely be improved by trying to guess the mime-type from the file content, instead of the file extension, but it is at least more secure than trusting the web browser.

protich:
Rejecting this implementation for now - as you mentioned it can be improved and there is no reason to accept it as it is now.

PR #197, opened 2012-08-02
Attachements can be downloaded AND displayed

soif:
When clicking on an attachement (from the Ticket view), it defaults to display the file in a new window (letting the browser display the file or not, depending on the mime-type). Files can still be (force) downloaded by clicking on the small icon.
IMO this is really more convenient, when most of the attachements are images or pdf.

(While this commit doesn't require my last "mime types" pull request, it should be more secure to include it.)

BTW I've currently harcoded a small style in the links to allow it to merge smoothly, in my current develop branch, without breaking my CSS branch. When you'd approve or deny my CSS pull request, I would move this to the exernals stylesheets, and make it more beautifull.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,3 +84,4 @@ osTicket is supported by several magical open source projects including:
   * [PEAR/Net_Socket](http://pear.php.net/package/Net_Socket)
   * [PEAR/Serivces_JSON](http://pear.php.net/package/Services_JSON)
   * [phplint](http://antirez.com/page/phplint.html) 
+


soif:
Latest commit adds "Include Content-Type: when available....to allow browsers to auto detect the mime type of the content stream." (issue #191)

protich:
The plan is to let admin decide (config setting) which mime types can be shown inline - within the ticket thread. The user's browser should be able to auto detect which mime types are viewable (hence issue #191) - based on available plugins like PDF viewer. Forcing a display when user's browser is not capable to display a content type is pointless.

Adding content type of the content stream is sufficient. I'll merge commit 240bf50 if you can rebase the pull request.

Thanks,

soif:
absolutey great. I was thinking about thatt too, but as I didn't knew how to propose a feature that includes sql updates, I did not investigate it further.

BTW, my first commit does NOT display attachement inlines: it just allow the user to either click to download the attachement (as now) but also click on another link that just open the attachement in a new window, without forcing the download. In this case, depending on the mime type, the browser should either display it (image, sound, pdf, etc...) either download it ...If you don't like the idea to have 2 links per attachement, IMO we should have only the non-forced download link, which is really more convenient in most case (ie images). The user can then easily right-click to download the attachement if he prefers... No?

PS: i don't understand what you mean by "rebase the pull request" (remember I'm a Git newbie) ;-)

protich:
I know your pull request doesn't include inline attachment view -I was just giving you the big picture. 

Yes - I don't like the idea of 2 links - adding mime type of the content stream is sufficient. 

By rebase - I meant redo your pull request to just 240bf50 commit.

soif:
Adding mime type of the content stream is -not- sufficient. The browser will still download the file whatever the mime type is: The other headers are forcing a download.

This is what you want? No ability to view a file (ie an image) without downloading it first ????

protich:
If your solution is to add 2 links - then I wholeheartedly reject it. 

The solution is to auto-detect _safe_ mime types (images/pdf...etc) and do a display on them instead of download. That needs to happen in attachment.php. My point is commit 240bf50 can be merged as it is while addressing display issue separately.

soif:
Surely some language incomprehension...

My solution is NOT to add 2 links: I suggest that -for now- (until we work on the inline-display feature), we use the class.file/display method instead of the class.file/download method for attachements. (I mean that I will commit changes to just have ONE link pointing to the display method, instead of the download method)

This would allow the user to easily  (just one click) display browser-safe mime type attachements, like images, without having to download it first, which is really a pain for him.

Then we should replace it with the inline view feature, which is of course the best one. But while waiting this features to be implemented (1.8? 2.0? another 2 years waiting?), IMO my solution seems really more user-friendly, and does not lead to major changes in the code ...

No?

protich:
That is exactly what I suggested - see my comments above  (you're the one who implemented 2 links solution in the first place). **My point is** instead of blindly using display - auto-detect _safe_ mime types first. For example you probably want to force a download on `.exe` file.

```
if($file->isSafeMimeType()) 
   $file->display();
else
   $file->download();

```

Get it?

soif:
Certainly again an incomprehension...

In my browser, if the mime-type is not safe, my browser will download it, instead of trying to display it.
Maybe some examples will make you better understand what I mean:

Lets say that we use the display instead of the download method, when the user click on an attachment link:
- the mime type is image/gif ---> my browser display the image in a new window
- the mime type is application/octet-stream or un-safe ---> my browser download it 

Obviously I don't checked what EVERY browser do when he is faced with a non safe mime type. I just supposed that he either download it (like in my case) or at least print it on screen as-is (garbage)...

In your example (.exe), my browser will get a "application/octet-stream" mime type, so he will download the file instead of trying to print it.

But anyway, 90% of attachements should be images (screenshots of an issue), so the browser will print it correctly. When this is not the case, if the browser is not like mine automatically downloading the attachment, he will just print garbage on screen, and the user is then just one click away to right-click on "Download this link as file".

I agree that this is still not as elegant as inline printing safe mime types, but in most cases, it should be more convenient for the user to display the attachment instead of downloading it every time first.

Did I miss something?

protich:
We both agree that images and some files like PDFs should be displayed while the rest need to be downloaded. Why show garbaged content to the user when we can auto-detect and force a clean download?  In addition, the download function in class.file.php adds extra headers like filename...etc

protich:
Are you seriously reading my comments or you just being disagreeable for the sake of it? 

I didn't say we need to do download 100% of the time - I **even** provided a pseudocode (2 comments above) on how we can auto-detect between display & download.

This is becoming a waste of time.

**Context:** This comment was in response/reply to @soif deleted comment quoted below.

> Absolutely Right, but why 100% of the time should we download the file, while 90% of time (let's say 75%) the file should be an >image, that should rather been displayed?
> 
> That perfectly resume my point of view ;-)

soif:
Absolutely sorry, but I when I replied , your pseudo code, was NOT displayed here... Some latter edit I guess....

(and sorry: I've deleted my previous comment, by mistake)

soif:
I changed it strictly  following your pseudocode. 

Are you finally happy with this pull request? :-)

soif:
Pull reseted to be mergeable into rc2.
The pull request now display the attachment when the mime type is safe, else file is still downloaded

Issue #198, opened 2012-08-02
Cancel on staff view


protich:
Clicking cancel on staff view redirects to departments - should go to staff members list.

PR #199, opened 2012-08-02
Issue/198

protich:
Fix Cancel button and Teams selection on staff page

Addresses issue #198 and #183


greezybacon:
Good catch. Thanks

Issue #193, opened 2012-08-01
Password change


protich:
Make sure the new password is not the same as the old password.

PR #200, opened 2012-08-02
issue/193

protich:
Make sure new password doesn't match current one - issue #193

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -408,6 +408,8 @@ class Staff {
                 $errors['cpasswd']='Current password required';
             elseif(!$this->check_passwd($vars['cpasswd']))
                 $errors['cpasswd']='Invalid current password!';
+            elseif(!strcasecmp($vars['passwd1'], $vars['cpasswd']))
+                $errors['passwd1']='New password MUST be different from the current password!';
         }
 
         if(!$vars['timezone_id'])


greezybacon:
I don't totally agree with this: since technically passwords never expire, there should be no real reason for the password to be forced to change. The only reason this comes about is because the encryption algorithm changed for osTicket 1.7. Would a better fix be to silently upgrade the password hash at the next time the staff logs in? ie. the code that checks if the password correct could detect that the password hash should be upgraded. Since the (correct) plain-text password is known at that time, the new hash could be calculated and updated in the database.

protich:
Upgrade was not the only reason! I can think of 2 reasons (that will totally change your mind) for forced password change.
- osTicket now ships with password change policy - it can be set in Admin Panel. 
- Admin can also manually force staff to change password - this comes in handy when the user forgets a password and admin simply resets it to "changeme" with forced password change on next login.

I still _totally_ stand by forced password change for reasons stated above. 

That sad, I decided against silent password change mainly because MD5 password doesn't necessarily mean the user hasn't logged in since the upgrade. MD5 based password can still be used to reset admin's forgotten password directly in the database. I now see possible flaws in my thinking... I shouldn't trouble everybody, on upgrade, just for possible future scenario.

greezybacon:
Agreed

PR #202, opened 2012-08-08
Saving recipient from email

Bastian-Kuhn:
I added a feature for saving also the recipient address of EMails to the Database.
There are a few benefits for us (may also for others)

We use customer@support EMail Addresses, so we can track now all tickets of a customer, it doesent matter from which employee  the mail was send.

We create the most Tickets EMail based, meaning we move the mail from private inbox to our osTicket inbox. With this change we doesent loose the originial recipient from the mail

I will add the frontend functions for it later (view, filter)

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -235,11 +235,13 @@ class MailFetcher {
             return null;
 
         $sender=$headerinfo->from[0];
+        $recipient=$headerinfo->to[0];
         //Just what we need...
-        $header=array('name'  =>@$sender->personal,
-                      'email' =>(strtolower($sender->mailbox).'@'.$sender->host),
-                      'subject'=>@$headerinfo->subject,
-                      'mid'    =>$headerinfo->message_id
+        $header=array('name'    =>@$sender->personal,
+                      'email'   =>(strtolower($sender->mailbox).'@'.$sender->host),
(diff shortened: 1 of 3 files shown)


protich:
I sort of see what you're trying to accomplish - but I'm interested in fully fleshing out the intent, so it (the feature) can makes sense for _everyone_ not just **your** internal use-case.
- Saving `recipient_email` is fine - but how exactly do you intent on using it?
- What is the difference between `recipient_email` and `emailId` (already supported)?
- How exactly do you move the emails - simply forward it? 

**PS**: Any changes that involve SQL/DB changes must support monkey-patching and also applied to installer. I realize we haven't documented the process yet - it might make sense to review a closed pull request with DB changes.

Bastian-Kuhn:
Saving: Case 1) Get all Tickets from a customer for billing. Case 2) Later: Assignment of Tickets by recipient instead of postbox Case 3) Employee moved ticket from his INBOX to the Ticket System, call back this colleague in case of questions. 

Difference: The emailID isnt the recipient address. It belongs to the originator of the mail

How move: No, not forwarding. Its done in postifx, he deliver the mail from multiple recipient addresses to a singe inbox.

greezybacon:
Would you mind reopening this in the 1.8 repo? We're strongly considering added it to osTicket 1.8.1

PR #223, opened 2012-08-13
Implement automatic client logins

greezybacon:
This commit allows for a link to be written in an email template

```
%url/view.php?e=%email&t=%ticket&a=%auth
```

which will short-circuit the client login procedure. The ticket-auth-token is generated using the ticketing system secret salt and the internal ticket id so that the auth-token is not externally guessable. As a convenience, `%clientlink` can be used in substitution of the above url to support changes to the client authentication system in the future. _ie. the link can be changed again in the future and no further changes will be necessary to the email templates._

@soif, what are your thoughts with this implementation?

### DIFF
--- a/include/ajax.content.php
+++ b/include/ajax.content.php
@@ -36,46 +36,48 @@ class ContentAjaxAPI extends AjaxController {
 
     function ticket_variables() {
 
-        $content='<div style="width:600px;">
-                    <h2>Ticket Variables</h2>
-                    Please note that non-base variables depends on the context of use.
-                    <br>
-                    <table width="100%" border="0" cellspacing=1 cellpadding=2>
-                        <tr><td width="50%" valign="top"><b>Base Variables</b></td><td><b>Other Variables</b></td></tr>
-                        <tr>
-                            <td width="50%" valign="top">
-                                <table width="100%" border="0" cellspacing=1 cellpadding=1>
-                                    <tr><td width="100">%id</td><td>Ticket ID (internal ID)</td></tr>
-                                    <tr><td>%ticket</td><td>Ticket number (external ID)</td></tr>
-                                    <tr><td>%email</td><td>Email address</td></tr>
-                                    <tr><td>%name</td><td>Full name</td></tr>
-                                    <tr><td>%subject</td><td>Subject</td></tr>
-                                    <tr><td>%topic</td><td>Help topic (web only)</td></tr>
-                                    <tr><td>%phone</td><td>Phone number | ext</td></tr>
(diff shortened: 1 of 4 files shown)


greezybacon:
This is a re-implementation of #181

soif:
I did not checked the details of your code , but here are my first thoughts.

1) The client auto-login feature  is definitively needed to provides the client with a smooth and friendly user-experience. So whatever the way it is implemented, it is way better than letting the customer in front of an "Authorization Required" form, when he had only clicked on the link we have emailed him. 

2) having a %clienlink is clever and will , as you said, let us change the auth later, if we want, without breaking existing templates But please also keep the %auth tag, to still allow finer template customization (ie different urls depending on the templates used)

3) It is not Bot proof : anyone can submit the ticket (authorized) link to google (as will any GoogleAds present on the ticket page automatically submit it to Google) 
Simply adding an expiration date of 1 month to the %auth link, should invalidate the link onces Google has grabbed it. Something like:
`If (date of_last_ticket_message < time() - 1 month) $auth==null;` 

In this case redirect the user to the login-form, and maybe use my simple javascript (#181) to auto-submit the login form. (Again, letting the customer in front of an unfriendly form, when we can avoid it to him, is bad!)

greezybacon:
@soif, yes, the %auth token is here to stay. Concerning bullet 3, if google indexed the current link system (with email and ticket), it would still have indexed access to the ticket view page. I don't exactly see how %auth changes the game.

soif:
great for the %auth

about Google:
If Google has recorded the url+%email+%ticket link, he has no access to the page (the form must be submitted). My Javascript was just clicking the submit form on behalf of the user. This should not happen with google (Google is not supposed to follow Javascript form submit... at this time)

If Google has recorded the url+%auth link, he will show the page on the next GoogleDance (3week to 1month, depending on the page rank). If the link is no longer valid once displayed on google (because auth has expired), it will prevent unauthorized accesses to the ticket page. 

adding a meta tag robots=noindex to the ticket view page, should also help, to prevent google to cache it.

protich:
@soif - I see your point on adding `noindex` to ticket view pages and may be even scp/*.

As for your google bot scenario - you're making the assumption that the user WILL somehow leak the **emailed** link to google. 
- Why would a user submit the **emailed** link to google?
- Having adsense on the page doesn't leak the link - it's an emailed link and page redirects on auto-login. Please clarify.

I'm not against adding some kind of timeout - but it **should** be a setting admin can enable/change.

protich:
Any further comments should be made in issue #227.

Issue #224, opened 2012-08-13
Sort ticket thread asc AND desc by date capabilities


jeffchulg:
When one wants to get directly to the last comment made on a ticket, he has to go to the end of the ticket thread. If there was a capability to sort ticket thread descending by date of creation, this problem would be solved. 
There also could be the possibility to set system/user default sorting preference.

soif:
would you like this way?

jeffchulg:
as you want, the principle is that you get access to the lastest messages easily

PR #172, opened 2012-07-26
CSS Enhancements

soif:
Fixes
- I've (auto) reformatted the CSS.
- Changed your comments, to CSSEdit friendly group comments
- Re groupped some styles
- Removed duplicates declarations

Features Added
- Started to improve the SCP look mainly
- Started to improve the Login window

If  you like, i can finish it (fix some table widths in the client view, and some margins)

Comments, like/dislike welcome!

PR #231, opened 2012-08-14
Auto Scroll to the latest message in ticket view

soif:
This auto scroll to the latest ( -1) message in the scp ticket view.
This is a proposal to solve issue #224 in a different way


soif:
#172

Issue #237, opened 2012-08-22
Timezone Issues


protich:
Some users are reporting timezone related issues. It happens when MySQL timezone comes back as "SYSTEM" - meaning local timezone. The offset in such cases should be zero.

slylenser:
I've found some issues related to timezone. 

The setting related to timezone in admin settings does not change the database config table. Despite this, staff panel tickets show the changed timezone while client panel shows the timezone based on database config table. Temp fix is to set the timezone in admin settings (say +5.5hrs for India) and then edit the config table in database and set timezone to 5.5 to match display in client and staff panels

klimnad:
## Warning: strtotime(): It is not safe to rely on the system's timezone settings. You are _required_ to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected 'Asia/Irkutsk' for 'IRKT/9.0/no DST' instead in /usr/share/doc/indexhtml/spravka/include/class.staff.php on line 61 

It is message i see when install the new clean - or upgrade. 
I solve this problem simply:
OR i was add in /include/class.config.php after line 51
date_default_timezone_set('Asia/Irkutsk');
Maybe it can be universal or automatic?

PR #238, opened 2012-08-23
Issue/237

protich:
Fixes time zone bug - issue #237


greezybacon:
Dude, I would prefer an admin setting so that the administrator can just declare which timezone the database is in. Perhaps you could make a guess for the instances where the administrator doesn't know which timezone the database is in (hosted databases), but I would suggest issuing a warning by the installer/upgrader if the database timezone if configured as 'SYSTEM'

protich:
I decided to go with auto-detect for now. I'll revisit timezones & dates post 1.7 release.

Issue #11, opened 2012-03-21
Knowledgebase Status


protich:
Set knowledgebase status to offline by default.

protich:
The change will require empty patch -> waiting on the next DB related change.

PR #251, opened 2012-08-31
Feature/filters revisited

protich:
Make filters generic and applicable to all incoming tickets.
Add ability to target the source (filters specific to source of ticket)


protich:
***\* Don't Merge Yet****

Pull request made so others may help test.

include/class.filter.php
```diff
@@ -208,35 +214,39 @@ function containsRule($what, $how, $val) {
...
      *   reply-to - reply-to email address
      *   reply-to-name - name of sender to reply-to
      *   headers - array of email headers
-     *   emailid - osTicket email id of recipient
+     *   emailId - osTicket system email id 
      */
-    function matches($email) {
+    function matches($info) {
+
+        if(!$info) return false;
```
greezybacon: [Like] button

Might wanna write

``` php
if (!$info || !is_array($info)) return false;
```

include/class.filter.php
```diff
@@ -279,13 +289,15 @@ function apply(&$ticket, $email=null) {
         #       XXX: Unset the other (of staffId or teamId) (?)
         if ($this->getStaffId())    $ticket['staffId']=$this->getStaffId();
         elseif ($this->getTeamId()) $ticket['teamId']=$this->getTeamId();
-        #       Override name with reply-to information from the EmailFilter
+        #       Override name with reply-to information from the TcicketFilter
         #       match
```
greezybacon, line 295: TicketFilter

include/class.filter.php
```diff
@@ -316,47 +328,55 @@ function update($vars,&$errors){
...
         }
 
         return $num;
     }
 
     /** static functions **/
-    function create($vars,&$errors){
+    function getTargets() {
+        return array(
+                'All' => 'All',
```
greezybacon: I would suggest

``` php
'Any' => 'Any'
```

To me, `All` implies more than one item, where `Any` is a wildcard

include/class.filter.php
```diff
@@ -693,45 +743,55 @@ function apply(&$ticket) {
...
         # sender-email-addresses or sender-names or subjects
         $sql.=") OR filter.id IN ("
                ." SELECT filter_id "
-               ." FROM ".EMAIL_FILTER_RULE_TABLE." rule"
-               ." INNER JOIN ".EMAIL_FILTER_TABLE." filter"
+               ." FROM ".FILTER_RULE_TABLE." rule"
+               ." INNER JOIN ".FILTER_TABLE." filter"
                ." ON (rule.filter_id=filter.id)"
+               ." WHERE filter.isactive"
+               ." AND (filter.target='All' OR  filter.target=".db_input($this->getTarget()).")"
```
greezybacon: This might look cleaner

``` sql
AND filter.target IN ('Any', "db_input($this->getTarget())")
```

greezybacon:
This implements issue #11 now with 7fea9f1

protich:
Jared - what is the review status on this PR? No rush merging - just want to make sure you're not waiting on me.

include/class.filter.php
```diff
@@ -208,35 +214,39 @@ function containsRule($what, $how, $val) {
...
+            'equal'     => array('strcmp', 0),
+            'not_equal' => array('strcmp', null, 0),
+            'contains'  => array('strpos', null, false),
+            'dn_contain'=> array('strpos', false)
         );
+
         $match = false;
         # Respect configured filter email-id
-        if ($this->getEmailId() && $this->getEmailId() != $email['emailId'])
+        if ($this->getEmailId() && $this->getEmailId() != $info['emailId'])
```
greezybacon: Need to also ensure

``` php
$this->getTarget() == 'Email'
```

Really, we should probably accept the `target` (`source` in `class.ticket.php::create` I believe) in the `$info` array and ensure that matches, but otherwise, the above block is a good start

greezybacon: Ok, I've had more time to think about this. I think this line needs to read something like

``` php
if ($this->getTarget() != 'Any' && strtoupper($info['source']) != strtoupper($this->getTarget())
    return false;

if ($this->getTarget() == 'Email' && $this->getEmailId()
        && $this->getEmailId() != $info['emailId'])
    return false;
```

Issue #222, opened 2012-08-13
%assigned_staff variable


protich:
- Change `%assigned_staff` variable to `%assigned`  
- Support team assignment 
- If ticket is assigned to both staff & team use **staff/team**

protich:
Planned variables.

`%assigned_staff` - Assigned staff
`%assigned_team` - Assigned team or may be just `%team`
`%assigned` - If ticket is assigned to both staff & team show staff/team - otherwise show staff or team.

protich:
Considering adding `%staff` (... and dropping `%assigned_staff`) as a base variable  - to represent assigned or closing staff. However, the change will break alerts and notices using `%staff` for staff name.

PR #253, opened 2012-09-02
Feature/variables revisited

protich:
Add `%team` and `%staff` to base variables
Add "object" level variables for staff, team and dept.

Do NOT merge yet: see issue #222 for pending discussion.


protich:
Doing a pull request for review purposes - I'll rebase it to chance the schema signature once attachment PR is merged.

include/class.variable.php
```diff
@@ -0,0 +1,145 @@
...
+
+        //Unknown object or variable - leavig it alone.
+        $this->setError('Unknown obj for "'.$var.'" tag ');
+        return false;
+    }
+
+    function _parse($text) {
+
+        $input = $text;
+        if(!preg_match_all('/'.$this->start_delim.'([A-Za-z\._]+)'.$this->end_delim.'/', $input, $result))
```
greezybacon: You don't have to escape the `.` inside brackets, and PHP allows number chars in the function names as well. I would suggest `[A-Za-z_][\w._]+`

include/class.ticket.php
```diff
@@ -1084,51 +1087,87 @@ function onOverdue($whine=true, $comments="") {
...
+            case 'phone_number':
+                return $this->getPhoneNumber();
+                break;
+            case 'auth_code':
+            case 'auth_token':
+                return $this->getAuthToken();
+                break;
+                
+            case 'clientlink':
+            case 'client_link':
```
greezybacon, line 1125: I really think we should be clear and simple with this. There should be only one way to do something. It's either `clientlink` or `client_link`, `auth_code` or `auth_token`, etc. I think here that flexibility  will just make things confusing. Personally, I like the all lower-case values with underscores

include/class.ticket.php
```diff
@@ -1216,8 +1252,8 @@ function transfer($deptId, $comments, $alert = true) {
             $sentlist=array();
             foreach( $recipients as $k=>$staff){
                 if(!is_object($staff) || !$staff->isAvailable() || in_array($staff->getEmail(),$sentlist)) continue;
-                $alert = str_replace("%staff",$staff->getFirstName(),$body);
-                $email->send($staff->getEmail(),$subj,$alert);
+                $alert = str_replace('%{recipient}',$staff->getFirstName(), $msg['body']);
```
greezybacon, line 1266: Why not something like

``` php
$alert = $this->replaceVars($msg['body'], array('recipient' => $staff));
```

and use `%{recipient.first_name}` in the body as the token?

include/class.ticket.php
```diff
@@ -1527,12 +1558,8 @@ function postNote($title,$note,$alert=true,$poster='') {
 
 
         if($tpl && ($msg=$tpl->getNoteAlertMsgTemplate()) && $email) {
-                    
-            $body=$this->replaceTemplateVars($msg['body']);
-            $subj=$this->replaceTemplateVars($msg['subj']);
-            $body = str_replace('%note',"$title\n\n$note",$body);
-            # TODO: Support a variable replacement of the staff writing the
```
greezybacon, line 1534: Why did you drop the TODO?

include/class.variable.php
```diff
@@ -0,0 +1,145 @@
+<?php
+/*********************************************************************
+    class.filter.php
```
greezybacon: Whoops

include/upgrader/sql/15b30765-7f35de42.patch.sql
```diff
@@ -0,0 +1,409 @@
+/*
+ * @version=1.6RC2+
```
greezybacon: 1.7RC2+

include/class.ticket.php
```diff
@@ -1527,12 +1558,8 @@ function postNote($title,$note,$alert=true,$poster='') {
...
-            #       note
+                   
+            $msg = $this->replaceVars($msg, array('note' => "$title\n\n$note"));
```
greezybacon: I vote for both `%{title}` and `%{note}` replacements, and just change the `%note` currently in the database to `%{title}\n\n%{note}`.

include/ajax.content.php
```diff
@@ -56,7 +56,9 @@ function ticket_variables() {
                     <tr><td>%status</td><td>Status</td></tr>
```
greezybacon: Why didn't you add the braces here?

include/class.dept.php
```diff
@@ -277,6 +281,23 @@ function delete() {
         return $num;
     }
 
+    //Replace staff's variables
+    function replaceVariables($text, $prefix='dept') {
```
greezybacon: This function doesn't appear to be used

include/class.staff.php
```diff
@@ -502,6 +506,26 @@ function delete() {
         return $num;
     }
 
+    //Replace staff's variables
+    function replaceVariables($text, $prefix='staff') {
```
greezybacon: This function doesn't appear to be used

include/class.team.php
```diff
@@ -169,13 +173,28 @@ function delete() {
         return true;
     }
 
+    //Replace staff's variables
+    function replaceVariables($text, $prefix='team') {
```
greezybacon: This function doesn't appear to be used

include/ajax.content.php
protich: It's planned - made the pull request for the initial review.

include/class.ticket.php
protich, line 1266: The plan (my private plan??) is to create a mailer class  ... e.g

``` php
$mailer = new osTicketMailer($email,....);
$mailer->send($recipients, $msg, $options);
```

The mailer will create a single instance of VariableReplacer and use it for every recipient.  In addition it will help abstract all the for loops used to fire alerts to staff.

Please note that recipient can be just an email address, staff object or client object - so probably not good idea to do `%{recipient.first_name}` in all templates. 

**PS**: Mailer class actually already exists - it's `class.mailer.php`. Subconscious mind at work!

include/class.variable.php
protich: I copied the copyright header - and forgot to change the name. Good catch.

include/class.ticket.php
protich, line 1534: Not exactly sure why - but I *think" my reasoning was - a note can be posted by system as part of an event log - adding `%{staff.name}` to internal note template will be problematic in such contexts.

include/class.ticket.php
protich: I agree with you - but _feel_ like it should be `%{note.title}` & `%{note.message}`. Which will give us the flexibility of doing `%{note.staff.name}` or `%{note.poster}` - resolving context issues when compared to `%{staff.name}` or even `%{poster}` for note poster. See my other comment on the same issue.

That said, it will open a pandora box on the need to add classes for note, message & response objects (which should extend something like class.threaditem.php). You get the idea... I simply chickened out.

I'll implement it _your way_ for now, and then we'll revisit variables post RC3 release. Please? Pretty please??

protich:
Jared - please review the PR when you get a moment.

greezybacon:
Peter, I don't see anything glaring. But at 1460+ line changes, it's really hard to feel totally comfortable. And what's more, this stuff is difficult and time consuming to test. We **really** need a redesign centered around test suites!

protich:
Most of the lines probably come from the SQL patch - I did extensive testing on the variables, a quick scan will do.

Issue #228, opened 2012-08-14
Dashboard > access permission to stats 


jeffchulg:
Hi, I was wondering if it's normal for a simple staff to get stats on other depts and other staff members stats. 

In my opinion, this can tell people "they never work"  by watching at it.. (there is a lot of people like that..)
- admins should have access to everything
- dept managers should have access to stats  corresponding to his dept and staff members (+ maybe when a teams has members from his dept)
- staff should have access only to his own stats and stats on teams he's member, and also global stats on its dept.

What is your opinion ?

PR #261, opened 2012-09-04
Fix several broken features of the dashboard reports

greezybacon:
Including (possibly) #228. 

Better implementation of the security on the staff tab is of course welcome


protich:
Jared - closing the pull request - Please review my changes in #299.

Issue #240, opened 2012-08-23
Attachments becoming corrupt at 1 MB filesize


josh1856:
imap fetching when image (PDF) are sent in via email file is becoming corrupt.

Error reads:
"Adobe Reader could not open 'filename.pdf' because it is either not supported file type or because the file has been damaged (for example, it was sent as an email attachment and wasn't correctly decoded)."

josh1856:
I have done some debugging and determined that regardless where the file is downloaded from (exchange/postfix/sendmail) the mail fetch is corrupting the files.  This is not happening to ALL files but only random which I have not been able to narrow down the specific issue.

josh1856:
I have now tried pop3 and imap pulling from multiple sources and confirmed that the bug exists in both.

josh1856:
The bug also occurs when you create a new ticket via the web portal.

greezybacon:
I bet it only happens with attachments over 100k

josh1856:
just tried 6 different files under 100kb and 6 over 100kb not all of them were corrupt over the 100kb

josh1856:
corruption appears to be happening at 1MB

greezybacon:
So my next guess is that AttachmentFile::save in the UPDATE statement, the second argument to the CONCAT function should be a BINARY string literal:

CONCAT(filedata, BINARY ' . db_input(...) . ')'

greezybacon:
include/class.file.php

Sorry I don't have a line number

josh1856:
what specifically needs to be changed?

greezybacon:
Yeah, but you need a space between BINARY and the quote character

josh1856:
like so?

```
            .' SET filedata = CONCAT(filedata, BINARY '.db_input($chunk) . ')'
```

greezybacon:
Yes 

Am Aug 24, 2012 um 9:41 schrieb josh1856 notifications@github.com:

josh1856:
Tried that and it doesn't appear to help with the 1MB files.

greezybacon:
Can you send me an attachment that is getting corrupted? (jared at osticket dot com)

josh1856:
I will send you one but it's pretty much any file above 1mb.

josh1856:
any updates on this?

josh1856:
I have verified that this issue also happens when you send an external (reply) to a ticket.

greezybacon:
[...]

PR #263, opened 2012-09-05
Fetch file data in chunks for downloads

greezybacon:
MySQL has a limit on the maximum amount that can be transferred in one statement. It's the `max_allowed_packet` setting. The value of this setting will be the approximate upper limit of attachments that can be handled by the database given the current access model for osTicket.

The issue came up for attachment uploads and was corrected, so that uploads are chunk inserted into the database. Downloads, however, were forgotten. Strangely, it took quite a bit of debugging to track down the problem.

This patch corrects attachment downloads by fetching 256kB chunks of the attachment at a time and sending them directly to the client. This will also overcome PHP's memory limit which would be the second-level blocker of
attachment sizes. Lastly, the `AttachmentFile::getData()` method is simulated using output buffering. This will provide the same access as the previous getData() method; however, it is still subject ot PHP's memory limits.

This addresses issue #240


josh1856:
I just implemented this and am still getting the same issue when it comes to binary data files.

josh1856:
It appears to be pulling data in random chunks however most of the files are either 216kb or 424kb and corrupting when I pull the file.

greezybacon:
So it looks like `max_allowed_packet` is more deceiving that I originally anticipated. I think I'm getting closer, please check my latest version out.

josh1856:
It did solve the packet problem as far as I can tell however it appears to not work correct with .xls files.  I am trying other types as well.  (FYI for testing attachments are set to .*)

greezybacon:
@josh1856, I have absolutely no idea why the process would discriminate XLS files. Could you recheck and mail me a copy of the XLS (sanitized) so I could witness and reverse the corruption?

greezybacon:
@josh1856, @protich this is the next iteration. This is a _major_ change to the attachment storage model including a new table to store attachment data in chunks, and upgrade script to transfer attachments currently in your database to the new model, and some monkey patching for the major upgrade path to prevent corruption of attachments into the %file table.

This patch is able to completely overcome MySQL's implied limits of the max_allowed_packet flag by storing chunks of 256kB into the %file_chunk table. We could potentially change the chunk size to be as large as 510kB or perhaps even 1022kB, but I really don't think it will make any real performance differences.

**Please beware when testing these commits.** The may destroy attachments in your database, and this patch cannot easily be rolled back.

**Please test a full install and a full upgrade from 1.6\* before merging**

**Do also note that the included upgrade script, as written, can only convert attachments up to 16MB.** Also note that this limit only applies to users of 1.7 already, not to the 1.6 upgraders, whose attachments will be converted with the attachment migrator and are only subject to PHP's memory limit.

protich:
**Why 16MB limit?** Is it because of the chunk sizes?

greezybacon:
Yeah, the upgrade script has to copy a max of n blocks, whatever that is. For now I wrote it at 64 blocks because I'm lazy. Supposing things work, I could write a script it to expand the upgrade SQL script to 512 (128MB) blocks or whatever.

protich:
Makes sense - but why do you care to chunk existing data?  Why not simply copy it to a chunk block and call it a day.

Also 256kB seems arbitrary to me - why not base it on max_allowed_packet? Let's say 80% with upper limit of 1Mb (MySQL max packet default). That said, what happens if the user migrates the database to different server with different settings? 1Mb max should cover it - but it's something worth discussing.

greezybacon:
To keep the block sizes even and predictable, I wanted to copy data from the current %file data to the new %file_chunk table in those predictable chunk sizes. Since I implemented that process in SQL, the block size there needs to match the block size in the software. Since I had previously used 256kB, I just stuck with it.

I would agree with 512kB block size, but I'm using the bin2hex() function to convert the binary data in the SQL string. This will double the size of the data, and will run over the 1MB max_allowed_packet. Therefore, with this approach (which I'm thinking is faster than db_input but haven't tested), the largest block size would be about 511kB.

I wrote the read() and write() methods so that the blocks don't have to be completely fetched from the database and/or buffered. I think this would work well if we wrap the AttachmentChunkedData in a PHP stream class which will allow us to use it as a php file. (See http://www.php.net/manual/en/function.stream-wrapper-register.php). Then you can interact with the files in the database as regular files -- for example

``` php
md5_file('db://135');

copy($file['tmp_name'], 'db://:new:');

fpassthru(fopen('db://135'));
```

It would let us make our coding much simpler and more streamlined for handling attachment data. It would require the `allow_fopen_wrapper` setting (ON by default), so I haven't mocked it up yet, but PHP 4.3.2 supports it.

Hope this clarifies. Maybe I'm off in left field...

protich:
After sleeping on it; I truly believe you're over-solving the original issue and in turn introducing unnecessary complexity and dependency. Too much smarts in my opinion - we should keep it simple and stupid (KISS) - sort of like file system's read & write operations.

**Chunked write solves all issues related to max_allowed_packet - no need to do "auto-chuncked" read.**
We absolutely don't need to have chunk sizes even or predictable - what matters is a working chunked insert. If it got inserted then we surely can read it. 

AttachmentChunckedData  should load chunked ids (`chuck_id` -  should be auto increment field) on class init.

``` php
$res = db_query("select chunked_id as id, LENGTH(filedata)as size  from CHUNKED_TABLE where file_id=$id");
while($row = db_fetch_array($res)) 
   $this->chunks[] = $row;

```

On read routine - you simply tie `$this->_pos` to index of `$this->chunks` and _boom_ you have read block/pos, chunk_id and size of the data. If need be (although it's not necessary at all), you can do chunked read on a chunk block (chunk read a chunk).

The write routine - should be as dumb as it gets. Accept data & chunk size and simply do split & insert. 

**No need to do bin2hex.**
I absolutely see no reason (I admit... I could be mistaken) to unpack file data and unnecessarily bloat the database. db_input is simply a wrapper for mysql_real_escape_string which is binary safe.

greezybacon:
I mostly agree with your read implantation. It fixes the upgrade issue; however, it does require an extra trip to the database.

You misunderstand the bin2hex bit though. There are two ways to encode string data: with real_escape, which will escape invalid chars like null and quote chars, etc. The other way is to encode the data in hex. I personally think the hex encoding method is cleaner and faster. It doesn't mean that the data is stored as hex in the database.

I'm in a pretty terrible mood and don't feel like defending my approach. You tell me how you want it and I'll rebase it.

josh1856:
Which version did we finally decide to merge into RC3?  9417be1?

greezybacon:
@josh1856, is your source code screwed up? Is there no `define` for `FILE_CHUNK_TABLE` set in `main.inc.php`? Your error worries me. How are you staging the code for testing? -- are you using patches or pulling my github branch? Did the system prompt you for an upgrade?

josh1856:
Jared,
Very odd…  we are pulling using github branch and the system did prompt for upgrade.

@josh1856https://github.com/josh1856, is your source code screwed up? Is there no define for FILE_CHUNK_TABLE set in main.inc.php? Your error worries me. How are you staging the code for testing? -- are you using patches or pulling my github branch? Did the system prompt you for an upgrade?

protich:
Upgrade works as expected. Did 1.7 RC1-2 upgrade as well as 1.6 ST.

josh1856:
I will try again then.

On Sep 26, 2012, at 10:19 PM, "Peter Rotich" <notifications@github.com<mailto:notifications@github.com>> wrote:

Upgrade works as expected. Did 1.7 RC1-2 upgrade as well as 1.6 ST.

josh1856:
The following solved the issue.  protich thanks for merging into the branch.

git stash
git checkout origin/develop
git reset --hard origin/develop

josh1856:
It appears that this issue has resurfaced...

https://github.com/osTicket/osTicket-1.7/issues/579

PR #276, opened 2012-09-17
Feature/notification bars

kellishaver:
There are sample notification bars at the top of inc/staff/header.inc.php that you'll want to remove and display dynamically.

### DIFF
--- a/scp/css/scp.css
+++ b/scp/css/scp.css
@@ -49,6 +49,12 @@ a {
 
 #msg_error { margin: 0; padding: 5px 10px 5px 36px; height: 16px; line-height: 16px; margin-bottom: 10px; border: 1px solid #a00; background: url('../images/icons/error.png') 10px 50% no-repeat #fff0f0; }
 
+#notice_bar { margin: 0; padding: 5px 10px 5px 36px; height: 16px; line-height: 16px; border: 1px solid #0a0; background: url('../images/icons/ok.png?1300763726') 10px 50% no-repeat #e0ffe0; }
+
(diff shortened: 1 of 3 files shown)


scp/css/scp.css
```diff
@@ -49,6 +49,12 @@ a {
 
 #msg_error { margin: 0; padding: 5px 10px 5px 36px; height: 16px; line-height: 16px; margin-bottom: 10px; border: 1px solid #a00; background: url('../images/icons/error.png') 10px 50% no-repeat #fff0f0; }
 
+#notice_bar { margin: 0; padding: 5px 10px 5px 36px; height: 16px; line-height: 16px; border: 1px solid #0a0; background: url('../images/icons/ok.png?1300763726') 10px 50% no-repeat #e0ffe0; }
+
+#warning_bar { margin: 0; padding: 5px 10px 5px 36px; height: 16px; line-height: 16px; border: 1px solid #f26522; background: url('../images/icons/alert.png?1307823786') 10px 50% no-repeat #ffffdd; }
```
greezybacon, line 54: Why do you need a timestamp on a static picture reference?

kellishaver, line 54: We don't. The COMPASS compiler adds it automatically. It's stilly, and it just never got stripped out... That said, the CSS is a bit of a mess in several places, so I'll finish switching SCP CSS over to LESS (we don't need all of the features of COMPASS and LESS is a little easier to deal with) and including the uncompiled source soon. I'll try and do it this weekend. I've been meaning to forever.

PR #277, opened 2012-09-17
Feature/ticket actions dropdown

kellishaver:
I went ahead and did the JavaScript, as well, because it seemed quicker than trying to explain to you what I was thinking with the data attributes and dialogs.

You'll need to edit some URLs in include/staff/ticket.view.php - they're marked.

I styled the select box with CSS to add icons. IE < 9 probably won't support them, but I figured this particular bit of decoration isn't significant enough to justify the overhead of a jQuery plug-in (file size and load time) just for older versions of IE.


kellishaver:
As we discussed, modifying this to expose the print and edit buttons next to the dropdown.

Print and Edit

protich:
Jared - The pull request was not meant to be merged yet -  I'll pull the branch locally to implement and resubmit it. Kelli  simply spec'ed the GUI for me.

PR #285, opened 2012-09-20
Little (not final) bandaid for 'umlaut' problems

Bastian-Kuhn:
There a currently again big problems with german "umlaute", ä ü ö, the subject ist cuted at the frist umlaute. With this pre workarround (mysql character set utf8) its a bit better, the umlautes stored as ? in the Database. This is better then a missing subject.  I currently dont have i final solution.


Bastian-Kuhn:
Mh, seems i'm wrong and this isnt just a bandaid, seems to be the solution. The umlauts in subject an from  field works now in my installation since this little change.

PR #289, opened 2012-09-26
Umlaut Problems

Bastian-Kuhn:
This now fixed the umlaut problems whit osTicket.

Not sure if the solution is in the way you want, but should be a base.

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -170,7 +170,7 @@ class AttachmentFile {
 
         foreach (str_split($file['data'], 1024*100) as $chunk) {
             $sql='UPDATE '.FILE_TABLE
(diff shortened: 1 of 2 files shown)


include/class.file.php
```diff
@@ -170,7 +170,7 @@ function save($file) {
 
         foreach (str_split($file['data'], 1024*100) as $chunk) {
             $sql='UPDATE '.FILE_TABLE
-                .' SET filedata = CONCAT(filedata,'.db_input($chunk).')'
+                .' SET filedata = CONCAT(filedata,"'.db_input($chunk, false, true).'")'
```
greezybacon, line 173: This was completely rewritten in #263, so this isn't an issue any more -- it uses hex encoding now rather than `db_input`

PR #292, opened 2012-10-02
added admin login (/scp) link to front page

markcunninghamgit:
added admin login (/scp) link to front page beside guest login
There needs to be a link to log into the admin interface.
There's no way in hell everyone's going to remember /scp after being away from the system for a while. 
If not here, then somewhere else

### DIFF
--- a/include/client/header.inc.php
+++ b/include/client/header.inc.php
@@ -33,7 +33,7 @@ header("Content-Type: text/html; charset=UTF-8\r\n");
                 <a href="<?php echo ROOT_PATH; ?>logout.php">Log Out</a>
              <?php
              }elseif($nav){ ?>
-                 Guest User - <a href="<?php echo ROOT_PATH; ?>login.php">Log In</a>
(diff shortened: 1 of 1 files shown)


koolaide:
Thanks for that idea! This will make my training much easier!

greezybacon:
I vote to make it more obscure and inconspicuous. I'd rather see it in the footer. Next to the copyright line or something

markcunninghamgit:
As long as there is a link somewhere.

ghost:
Can this be added as an option in the config? I for instance will be keeping /scp accessible only via a staff restricted domain, while  denying access to it on client public domain using Apache rules. So for me it would not be suitable to have the scp login link on public page. Customization for different situations should be considered, just my two cents. =o)

ntozier:
If this was to be added I would rather see this as a configuration option that could be toggled in say Admin panel -> System Settings & Preferences -> General Settings -> [] Show admin login link to clients.

Considering how easy it would be for someone to implement this change themselves it might not be worth adding it to core.

PR #296, opened 2012-10-05
Feature/canned auto response

protich:
- Fix canned auto-response 
- Filters related enhancements.

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -677,7 +677,8 @@ class TicketFilter {
      * return immediately.
      */
     function getMatchingFilterList() {
-        if (!isset($this->short_list)) {
+
+        if (!$this->short_list) {
             $this->short_list = array();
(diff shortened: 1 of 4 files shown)


include/class.filter.php
```diff
@@ -677,7 +680,8 @@ function getTarget() {
      * return immediately.
      */
     function getMatchingFilterList() {
-        if (!isset($this->short_list)) {
+
+        if (!$this->short_list) {
```
greezybacon: This will cause `getMatchingFilterList` to rematch all the filters every time it is called if there are no matching filters. The `isset` function cal was intended to handle that case

protich: Try ...

``` php
$a =array();
echo isset($a)?'YES':'NO';

```

PS:  short_list is getting cleared on every build or rebuild - see `build()`

PR #337, opened 2012-10-30
Allow event to trigger hook to customize further osticket 1.7

AlbanSeurat:
I need specific code to be use within the process of a ticket creation or many other event. I introducing a event mecanism in osticket to allow anyone to do custom code easily 

(inspired both by wordpress architecture and classical event design pattern)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,9 @@ php53.cgi
 include/ost-config.php
 *.sw[a-z]
 .DS_Store
+.AppleDouble
 .vagrant
 Vagrantfile
+include/hooks/*
+!include/hooks/hook.logdata.php
+!include/hooks/hook.mailsend.php
--- a/include/class.event.php
+++ b/include/class.event.php
@@ -1,0 +1,131 @@
+<?php 
+/*********************************************************************
+    class.email.php
+
+    Alban Seurat <alkpone@alkpone.com>
+    http://www.albanseurat.com/
+
+    Released under the GNU General Public License WITHOUT ANY WARRANTY.
+    See LICENSE.TXT for details.
+
+    vim: expandtab sw=4 ts=4 sts=4:
+**********************************************************************/
+
+class Event 
+{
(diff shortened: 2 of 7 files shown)


protich:
Love it!

My quick reaction is - may be it's worth thinking about events in context of plugins. In other words, plugins can hook to events which in turn are triggered by core osTicket. Just a thought.

AlbanSeurat:
You can indeed introduce a concept of plugin on top of a suite of hooks. I use a single hook file as a "plugin" myself. 
It can be futher improved with time

andrewtite:
Can you provide a code example for a basic plugin using this codehook? I'm trying to figure out a way to link into the receive and send functions so I can integrate eBay API Messages.

greezybacon:
@Alkpone, we opted to go for a very simple "signal" interface for extensibility. I feel like it's more anonymous and will play better so that extensions don't have to register their event codes in the mainline code. I'd love to hear your comments

https://github.com/osTicket/osTicket-1.7/blob/develop/setup/doc/signals.md

AlbanSeurat:
I don't get how an extension "script/file/class" is called from osticket. Somewhere you will need to have something calling "Signal:Connect" ?

How to you do that ? 

Look at the latest improvement made (not put in my pull request) about event (which is quite similar to your signal i guess) : 

https://github.com/Alkpone/osTicket-1.7/blob/develop/include/class.event.php

more specialized component "Plugin" which can register multiple events from a same plugin:

https://github.com/Alkpone/osTicket-1.7/blob/develop/include/plugins/plugin.highpriority.php
https://github.com/Alkpone/osTicket-1.7/blob/develop/include/class.plugin.php

greezybacon:
My concept is for plugins to have a bootstrap() function which would be used to do the Signal::connect() calls and plumb the signals originating from outside the plugin to the code the plugin brings to the party

AlbanSeurat:
Looks very similar to my events / plugin architecture

I can only appreciate your approach :)

Issue #322, opened 2012-10-19
"Starts With" in ticket filter rules


nekvindao:
It would be great if there was a "starts with" operator in ticket filtering rules.

Is something like this planned?

Thank you.

PR #378, opened 2012-11-16
Add support for starts- and ends-with in filter rules

greezybacon:
Match text at the beginning and ending of ticket fields


greezybacon:
Implements issue #322

PR #392, opened 2012-11-26
Adding additional function to have special reply when ticket is closing

JanKraljic:
It is very nice that reply of email is little different when the ticket is closing.

I have added addition attributes to mail template table and change code little so take other msg template.


protich:
Thanks for the pull request. Couple of comments....
1. Not sure I totally buy into separate template in place of the reply template. We've internally discussed possibility of configurable automated followup email - after the ticket is closed. Let's revisit the idea post 1.7 release. 
   FYI: Making DB changes also requires upgrade (fast forward) patch - see a closed pull request that involved db changes as a guide.
2. Make d7655d9 a separate pull request. 
3. 581810d is completely unnecessary - osTicket already does it. See `onMessage()`.

JanKraljic:
Mostly i'm preparing customs for our usage. I someone else find it usefull so bi it.

at 1: as you find it is best for your directions

at 2. done

at 3: i'll defenitly check it if it is working as it should and do revert

PR #394, opened 2012-11-27
Changed user GUI to change priority of already submitted ticket.

JanKraljic:
This is practical when problem is getting worste.

Just seperate pull request


JanKraljic:
Sorry i dont knot how properly create pull request for just one commit

PR #399, opened 2012-11-30
Fix redirect bug in l.php for IE8

Hamsteren:
Fixes bug in IE 8 32bit browser that makes the browser windows reload in a never ending loop.


protich:
Tommy - Thank you for the fix. Please apply the same fix to `l.php` on the base directory and we'll merge the pull request.

Hamsteren:
I have added the fix to l.php in the base directory now also.

PR #403, opened 2012-12-05
Better support for international charsets

kopitoshka:
Better support for international charsets.

The original code does not correctly convert mime lines like this one:
string(47) "=?windows-1251?B?0e3u4uD/IP8uIMPk5SDv7uzu+fw/?="

It results in an empty subject of a new ticket submitted via email.

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -206,14 +206,22 @@ class MailFetcher {
         return utf8_encode($text);
     }
     
-    //Generic decoder - mirrors imap_utf8
+    // The function returns a text in utf-8 charset always
+    //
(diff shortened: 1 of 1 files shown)


protich:
Thank you for the pull request. Your solutions sort of works, but ONLY if you have `iconv_mime_decode` and string/text doesn't contain utf-8 encoded parts. 

See commit 449e561 on how I went about solving the issue. Your input will be appreciated.

kopitoshka:
> See commit 449e561 on how I went about solving the issue. Your input will be appreciated.

It works fine for me. Thank you.

protich:
Addressed in pull request #406.

Issue #404, opened 2012-12-05
Inline attachments import issues.


protich:
Problems parsing file name when multiple disposition attributes exists.

PR #406, opened 2012-12-06
Attachments handling

protich:
- Fine correct filename when a disposition as multiple parameters
- Encode & remove white spaces on filenames
- Upload validation improvements


protich:
Addresses issue #404.

PR #410, opened 2012-12-07
Implemented MS-TNEF file handling in e-mail import feature

fabiorauber:
Conflicts:

```
include/class.mailfetch.php
```


greezybacon:
@fablorauber, can you provide a sample email or two or provide a link to where I could get some attachments to try out against this patch?

ntozier:
You should be able to create an email to test this patch by using Outlook.  You would want to configure Outlook by going to tools -> options -> Mail Format tab, and changing the "Outlook Rich Text options" to "Send using Outlook Rich Text Format" and sending an email with an attachment.  If you dont have Outlook or access to it, PM me an email address to send an email to and I will. :)

I have to say though that I'm not really in favor of adding a patch to support a single email client some of the time.

greezybacon:
This is implemented in osTicket/osTicket-1.8#555 (released in v1.8.0.3 and v1.8.1)

PR #428, opened 2012-12-19
Fix redirect &(ampersands) error.

Hamsteren:
Fix issue with &(ampersands) in url's getting parsed.
Url's where getting cut at the first ampersand.

### DIFF
--- a/l.php
+++ b/l.php
@@ -14,7 +14,9 @@
     vim: expandtab sw=4 ts=4 sts=4:
 **********************************************************************/
 require 'secure.inc.php';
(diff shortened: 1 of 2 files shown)


protich:
Thank you for the find & the pull request.

The fix didn't really address the root cause of the issue - which was unencoded url. See 7a2768a on how I went about resolving the issue. Your feedback will be appreciated.

Hamsteren:
Thank you for the fix, i have just tested this and it works like a charm.

And i would never have fixed the root cause myself as i'm not a programmer :)

PR #444, opened 2013-01-10
﻿Added support to In-Reply-To and References headers for sent and received ticket email replies.

lukakama:
Actual changes:
- For every "ticket reply alert" sent to an user:
  - It is saved a record inside the  "ticket_email_info" table with the messageId, linking it to the originating support ticket reply
  - In-Reply-To and Refecerences headers are added to the sent mail, both set to the messageId of the latest email received from the user (if available), in order to help user email client to properly thread the ticket discussion.
- For every incoming email, In-Reply-To header, Refecerences header and the sender email are used to lookup the existing ticket. If found, it is used as ticket where to append the reply, otherwise the standard subject ticket id lookup is performed.

That's all. It seems to work nicely, and allows to omit the ticket ID from the mail subject of support replies (as the subject alteration break gmail email threading).

I tested it with IMAP access, so I don't know if it works correctly with POP3.

Some things to review:
- I don't know if it is correct to use the "References" header for the lookup, because it seems that it can have multiple message ids inside it. Maybe it is to split-up in order to lookup each id one at time.
- Actually, I converted the email headers sent to the user to the text saved inside the "email info" table with a print_f function. I'm pretty sure that it needs to be improved to match the format of headers saved for incoming email :)

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -108,7 +108,14 @@ class Mailer {
                 'Message-ID' => $messageId,
                 'X-Mailer' =>'osTicket Mailer'
                );
-
+               
+        // Set In-Reply-To and References email headers to the latest email message when receiving the ticket_id.
+        if ($options && $options['ticket_id'] && ($ticket_mid = Ticket::getLastEmailMessageId($options['ticket_id']))) {
+            $headers+= array(
+                    'In-Reply-To' => $ticket_mid,
+                    'References' => $ticket_mid);
+        }
+        
         //Set bulk/auto-response headers.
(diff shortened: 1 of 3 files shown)


protich:
Thanks for the contribution. Jared (@greezybacon) already has a pending pull request addressing the same issue. I'll let him take the lead on reviewing your pull request.

Some comments;
- We definitely need to split and lookup on each reference ids starting from the latest.
- Outgoing emails ID needs to be saved in email_mid field and email_info table should reference thread_id
- CC/BCC needs to be considered in how outgoing message ids are stored (possible collisions...etc)
- Lookup should be on reference id + email or ticket id.

PR #454, opened 2013-02-09
Canned responses show up in menu when disabled

greezybacon:
Disabled canned responses still show up in the canned responses menu even if they are disabled. They are clickable; the only difference is the response doesn't get added to the email. The disabled responses should be either greyed out or (my personal preference) just not show up in the menu at all.

### DIFF
--- a/include/class.canned.php
+++ b/include/class.canned.php
@@ -217,11 +217,13 @@ class Canned {
 
     function getCannedResponses($deptId=0, $explicit=false) {
 
-        $sql='SELECT canned_id, title FROM '.CANNED_TABLE;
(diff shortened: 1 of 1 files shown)


greezybacon:
@protich do you feel like this is the right approach?

protich:
@greezybacon - Yes for now. That said, it make sense to add getActiveCannedResponses routine as a wrapper.

``` php
getActiveCannedResponses($deptId=0, $explicit=false) {
return self::getCannedResponses(array('deptId'=>$deptId, 'explicit'=>$explicit));
} 

```

then do -  `getCannedResponses($criteria)`

But on a second thought it can wait until we do ORM - where $criteria will be a class.

PR #473, opened 2013-02-20
check max file size at client before sent to server

fanha99:
when try to upload a file which is bigger than the max file size, there will be many error "invalid file type" in the ticket thread.

this is not best solution, since the problem is in the server side code.


fanha99:
hi, really sorry,
I'm using windows, and my stupid editor convert all the EOL to windows style.

I recommit the other branch with name "check_file_size".

could you please check it?

protich:
Resubmit the pull request on the other branch. Thanks.

PR #482, opened 2013-02-27
Added ability to show/hide department column without assigned column aff...

henesnarfel:
We need the ability to show both the assigned to and department columns on the open ticket list.  I've added another config item the same as 'show assigned tickets'.


henesnarfel:
Closing as this change is performing actions that the settings weren't meant for.

PR #485, opened 2013-03-01
Truly allows admins/managers to hide/show assignedto/department columns

henesnarfel:
This builds on issue #482 where it adds the ability to show/hide the department column in open tickets.

This fix allows admins/managers to truly hide the assignedto/department column by their preferences.  The scenario it fixes is below
- Before
  - Global Assigned To: ON
  - Personal Pref.: OFF
  - Column Still showed
- After
  - Global Assigned To: ON
  - Personal Pref.: OFF
  - Column no longer shows


protich:
James,

Thank you for the pull request. It's a good idea, but I don't necessarily agree on the goal as implemented. 

Let me explain;

Ability to show assigned tickets (on open queue) is different from ability to show Assigned To column. Hiding Assigned To column on open queue is sort of a side-effect of the later i.e _it's useless to show the column when displaying just unassigned tickets_. I know it's a little confusing - but it was never meant as away to hide/show Assigned To column. See pending pull request #477 for a little more insight on ticket queue & Assigned To column.

The correct way, in my opinion, to show/hide columns is to create tickets listing templates for queues. Basically away for admin to pick which columns to show for a specific queue/view (open, answered, overdue, closed...etc). Doing it on individual columns, as you implemented, is not sustainable since you'll have to add setting for each column on a global basis and may be on queue level too. Add dynamic fields (upcoming feature) to the mix and it becomes a nightmare.

I look forward to your thoughts.

henesnarfel:
Sorry for the late response.  But yes after your explanation I understand what those settings are meant for.  I agree that the admin should be able to decide what columns should and shouldn't be shown.

Issue #395, opened 2012-11-27
Link to FAQ in ticket view


jeffchulg:
Would it be possible to put a link depending on the help topic which would redirect to the list of faqs for that help topic (and its nested help topics) in 1.7 release ?

jeffchulg:
Here is the code to put in "include/staff/ticket-view.inc.php"  (starting at line 238):

```
            <tr>
                <th width="100">Help Topic:</th>
                <td>
                    <a href="/scp/kb.php?a=search&q=&cid=&topicId=<?php echo $ticket->getTopicId()?>">
                        <?php echo Format::htmlchars($ticket->getTopic()->getName());?>
                    </a>
                </td>
            </tr>
```

An alternative to get topic's name could be to change "getHelpTopic()" function of the Ticket class to do $ticket->getTopic()->getName()); in a good way

PR #507, opened 2013-03-14
Add link to help topic associated FAQs

jeffchulg:
Addresses #395

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -250,8 +250,10 @@ if($ticket->isOverdue())
         <td width="50%">
             <table cellspacing="0" cellpadding="4" width="100%" border="0">
                 <tr>
-                    <th width="100">Help Topic:</th>
-                    <td><?php echo Format::htmlchars($ticket->getHelpTopic()); ?></td>
+                    <th width="100">Help Topic:</th>                    
+                    <td>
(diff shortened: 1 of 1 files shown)


include/staff/ticket-view.inc.php
```diff
@@ -250,8 +250,10 @@
         <td width="50%">
             <table cellspacing="0" cellpadding="4" width="100%" border="0">
                 <tr>
-                    <th width="100">Help Topic:</th>
-                    <td><?php echo Format::htmlchars($ticket->getHelpTopic()); ?></td>
+                    <th width="100">Help Topic:</th>                    
+                    <td>
+                     <a href="/scp/kb.php?a=search&q=&cid=&topicId=<?php echo $ticket->getTopicId()?>">
```
greezybacon: I think this should be

``` html+php
<a href="<?php echo sprintf("%s/scp/kb.php?a=search&topicId=%s", $cfg->getBaseUrl(), $ticket->getTopicId()); ?>">
```

So that the installed location (URL base) of the ticketing system is not lost

jeffchulg:
Ok, I'll change it

greezybacon:
@jeffchulg, would it make more sense to link to a search based on the help topic?

jeffchulg:
As you want, my point is that if a FAQ is actually a procedure, it's easy for us to just click on the link provided to get access to the corresponding procedure.

Issue #510, opened 2013-03-19
wrong function call in class.ticket.php: onOverdue()


belzael:
There is a wrong function call in the file class.ticket.php on line 972:
            if($this->isAssigned() && $cfg->alertAssignedONTransfer()) {
it should be:
            if($this->isAssigned() && $cfg->alertAssignedONOverdueTicket()) {

greezybacon:
Could you review this with the changes merged in the develop branch?

belzael:
Looks alright to me.

PR #514, opened 2013-03-21
Fix a method name typo

protich:
Addresses issue #510.


greezybacon:
Drop the `$groupby` from `include/ajax.tickets.php:179`

greezybacon:
Drop the `$this->` in the `$this->die()` call in `setup/cli/modules/unpack.php:121`

PR #518, opened 2013-03-28
Overdue flag on ticket update.

protich:
Clear overdue flag if duedate or SLA changes results in new est. due date in the future.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1651,6 +1651,14 @@ class Ticket {
         $this->logNote('Ticket Updated', $vars['note'], $thisstaff);
         $this->reload();
 
+        //Clear overdue flag if duedate or SLA changes and the ticket is no longer overdue.
+        if($this->isOverdue()
+                && (!$this->getEstDueDate() //Duedate + SLA cleared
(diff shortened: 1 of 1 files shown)


include/class.ticket.php
```diff
@@ -1651,6 +1651,14 @@ function update($vars, &$errors) {
         $this->logNote('Ticket Updated', $vars['note'], $thisstaff);
         $this->reload();
 
+        //Clear overdue flag if duedate or SLA changes and the ticket is no longer overdue.
+        if($this->isOverdue()
+                && (!$this->getEstDueDate() //Duedate + SLA cleared
+                    || Misc::db2gmtime($this->getEstDueDate()) > Misc::gmtime() //New due date in the future.
+                    )) {
+            $this->clearOverdue();
```
greezybacon, line 1659: Perhaps just a caveat, but `class.ticket.php:1105` where the SLA is cleared (if due date is in the past) does not use the same date arithmetic as you are here. Is there a possible edge case where `$this->clearOverdue()` would clear the SLA?

protich, line 1659: Good catch -  I'll use the same GMT arithmetic.

Issue #519, opened 2013-03-29
Email Name Required when piping


bensuthers:
If an email is sent to the API via external pipe that doesn't have a 'name', eg just an email address, the following error occurs. 

API Error (400)
Unable to create new ticket: validation errors:
name: Name required
err: Missing or invalid data - check the errors and try again
_[A7A3358288FAD9976DCC5F77FCD4AA09]_

protich:
Thanks for the bug report - feel free to test above pending pull request or f32b35d.

bensuthers:
Hey Peter,

Thanks.. I'm not familiar with the codebase, but that hasn't fixed my issue. 

I think, that maybe  you've fixed it for email_fetch... but I was piping an email in using automail,pl from an external server....

Thanks

Ben

protich:
Ben - Are you sure you replaced both files? In your case it's class.mailparse.php

bensuthers:
Thanks Peter! Of course, I hadn't. I didn't see the other file to replace.

It's all working fine now. Thank you very much.

Bug fixed!

protich:
Thanks for the update - now I can go to sleep :) We'll get the fix merged to develop branch after a review.

PR #520, opened 2013-03-29
Use email address when FROM field doesn't have a name.

protich:
Addresses issue #519.

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -372,16 +372,20 @@ class MailFetcher {
 
         $emailId = $this->getEmailId();
         $vars = array();
-        $vars['name']=$this->mime_decode($mailinfo['name']);
         $vars['email']=$mailinfo['email'];
+        $vars['name']=$this->mime_decode($mailinfo['name']);
         $vars['subject']=$mailinfo['subject']?$this->mime_decode($mailinfo['subject']):'[No Subject]';
         $vars['message']=Format::stripEmptyLines($this->getBody($mid));
         $vars['header']=$this->getHeader($mid);
         $vars['emailId']=$emailId?$emailId:$ost->getConfig()->getDefaultEmailId(); //ok to default?
-        $vars['name']=$vars['name']?$vars['name']:$vars['email']; //No name? use email
         $vars['mid']=$mailinfo['mid'];
 
-        if(!$vars['message']) //An email with just attachments can have empty body.
+        //Missing FROM name  - use email address.
+        if(!$vars['name'])
+            $vars['name'] = $vars['email'];
+
+        //An email with just attachments can have empty body.
+        if(!$vars['message'])
             $vars['message'] = '(EMPTY)';
 
         if($ost->getConfig()->useEmailPriority())
--- a/include/class.mailparse.php
+++ b/include/class.mailparse.php
@@ -289,11 +289,15 @@ class EmailDataParser {
                 break;
             }
 
+            $data['email'] = $from->mailbox.'@'.$from->host;
+
             $data['name'] = trim($from->personal,'"');
             if($from->comment && $from->comment[0])
                 $data['name'].= ' ('.$from->comment[0].')';
 
-            $data['email'] = $from->mailbox.'@'.$from->host;
+            //Use email address as name  when FROM address doesn't  have a name.
+            if(!$data['name'] && $data['email'])
+                $data['name'] = $data['email'];
         }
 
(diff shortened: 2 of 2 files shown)


include/class.mailfetch.php
```diff
@@ -372,16 +372,20 @@ function createTicket($mid) {
...
         $vars['message']=Format::stripEmptyLines($this->getBody($mid));
         $vars['header']=$this->getHeader($mid);
         $vars['emailId']=$emailId?$emailId:$ost->getConfig()->getDefaultEmailId(); //ok to default?
-        $vars['name']=$vars['name']?$vars['name']:$vars['email']; //No name? use email
         $vars['mid']=$mailinfo['mid'];
 
-        if(!$vars['message']) //An email with just attachments can have empty body.
+        //Missing FROM name  - use email address.
+        if(!$vars['name'])
+            $vars['name'] = $vars['email'];
```
greezybacon, line 385: How is this different from original line 381?

greezybacon:
Personally, I feel like this should be changed in the validation in the `api.tickets.php` rather than monkey patching the mail parsing equipment. Technically, if an email doesn't have a named from address, then it doesn't have a named from address.

In other words, to me, it seems to raise the question if the 'name' field is really required. It seems we're working around incorrect requirements rather than making sense of the input.

include/class.mailfetch.php
protich, line 385: Just moved it down to group what we're monkey patching together - e.g empty message.

protich:
@greezybacon - I agree with you to some extend - but the name patch is specific to email, not all other formats (xml/json) or even web need it. We have to be **very** forgiving when it comes to emails - e.g empty message.

greezybacon:
I hear what you're saying, but why are creating two or more faces to the ticketing system? If an empty message is allowed via email but not via the web interface (staff or client), what are really requiring? We're truly just penalizing folks who submit tickets via the web. And, as a user, if I get fed up with the web interface, I'll just email in crappy emails with no name or message, and the ticketing system is suddenly just fine with missing data.

It's the same system. Why does the mode of input suddenly dictate what data is quote "required" by the system. I personally would vote for a return email stating validation errors to the email sender or make the same data interpretations on the web interface, if the name can technically dup the email address, and a ticket can be created without a message.

protich:
No name email or empty message is not crappy email - it's allowed as per email specs. Should email specs dictate how we develop the rest of the software? I would say **no**. 

Email is just not a good medium to return errors - given how SMTP works. If we're going to accept emails then we have to be liberal - no way around it - good example is OutLook formatting.

greezybacon:
Please understand, I see the issue, and I (would like to think I) understand your argument.

What I'm saying is this: (with a submission to the ticketing system with no name and no message): 
- via the staff web interface: **fail**: required fields not given
- via the client web interface: **fail**: required fields not given
- via the HTTP API: **fail**: required fields not given
- via email: no problem. Message becomes `(EMPTY)` and name dups the email field

What I'm suggesting is that the email interface mimic the rest of the interfaces. A return email could/should be sent indicating the ticketing system's refusal to create a ticket based on the email. The same process and feedback that would be represented via the other interfaces. This would be like a bounce email from the email system when the destination mailbox cannot be reached.

What I'm trying to convey is that the interfaces be consistent. If a message is not required, then we should just support creating a ticket with just a subject and an empty thread. If the name field is not truly required, then we should make it optional.

-- That's all I'm saying --

_PS_ - Dynamic forms worsen this issue, because they add even more fields that cannot necessarily be represented in MIME headers.

protich:
I understand what you are saying and the overall goal. I'll be totally with you, if osTicket was M2M application. 

We'll have to just agree that not ALL interfaces are made equal in this case. Should the emailing end user even care or know that an organization is using osTicket in the backend? 

The difference, in my opinion, is - admin/installer can/should controls web forms and API (json/xml) interfaces - but not email delivery. Bouncing "deliverable" emails back to the **end users** - just because it's missing a name or message is just way too much - 1) osTicket doesn't control MTA to cleanly bounce incase of IMAP/POP 2) Emailing end user doesn't have ability to correct and send the email - the mail server has already accepted the email. 3) The said errors have nothing to do with email delivery.

Dynamic forms/fields illustrates my view point - you might as well not add extra fields if email interface will dictate what we allow. 

PS: Think of end users as typical mom and pap shop customers.
PSS: It also helps to think of email integration as a non-core optional supplemental interface

PR #524, opened 2013-03-30
cli package revisited.

protich:
Package the correct scp sub-directories  -> css images js
Exclude ost-config.php
Use the correct package name

### DIFF
--- a/setup/cli/package.php
+++ b/setup/cli/package.php
@@ -89,6 +89,9 @@ package("*.php", 'upload/');
 foreach (array('assets','css','images','js') as $dir)
     package("$dir/*", "upload/$dir", -1, "*less");
 
+# Load API
+package('api/{,.}*', 'upload/api');
+
 # Load the knowledgebase
 package("kb/*.php", "upload/kb");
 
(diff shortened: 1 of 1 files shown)


setup/cli/package.php
```diff
@@ -89,22 +89,23 @@ function package($pattern, $destination, $recurse=false, $exclude=false) {
...
 
 # Load in the scripts
 mkdir("$stage_path/scripts/");
 package("setup/scripts/*", "scripts/", -1, "*stage");
 
 # Load the heart of the system
-package("include/*.php", "upload/include", -1);
-# And the sql patches
-package("include/upgrader/*.sql", "upload/include/upgrader", -1);
+package("include/{,.}*", "upload/include", -1, array('*ost-config.php', '*.sw[a-z]'));
```
greezybacon, line 108: What's up with the commas? Do we have files with commas in the filenames?

protich, line 108: Separator - basically means "any, ." - glob doesn't like spaces so you can't do " , ."

greezybacon:
Have you done a complete install via FTP with the ZIP package emitted?

protich:
I did yesterday and it didn't go well - hence the fixes here. I'll be doing some more testing today.

diff'ing stage/upload with the live code helped - we should probably add it to packager.

Issue #525, opened 2013-03-31
Closed tickets count.


protich:
Closed tickets count on ticket's sub-nav is wrong - it's showing "My Closed Tickets" count instead of all closed tickets.

PR #526, opened 2013-03-31
Fix closed tickets stats count

protich:
Addresses issue #525

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1764,8 +1764,7 @@ class Ticket {
                         AND assigned.staff_id='.db_input($staff->getId()).')'
             .' LEFT JOIN '.TICKET_TABLE.' closed
                 ON (closed.ticket_id=ticket.ticket_id
-                        AND closed.status=\'closed\'
-                        AND closed.staff_id='.db_input($staff->getId()).')'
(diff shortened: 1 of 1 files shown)


include/class.ticket.php
```diff
@@ -1764,8 +1764,7 @@ function getStaffStats($staff) {
                         AND assigned.staff_id='.db_input($staff->getId()).')'
             .' LEFT JOIN '.TICKET_TABLE.' closed
                 ON (closed.ticket_id=ticket.ticket_id
-                        AND closed.status=\'closed\'
-                        AND closed.staff_id='.db_input($staff->getId()).')'
+                        AND closed.status=\'closed\' )'
```
greezybacon, line 1767: So help me out. In the function `getStaffStats()` we're giving a staff member credit only if they currently own a closed ticket (ie. someone else could have actually closed it)? Should we transition to scanning the history table for staff stats?

protich, line 1767: These are not staff's historical stats - but live ticket counts based on staff's access.

greezybacon, line 1767: Closed is a realtime stat?

protich, line 1767: Basically the query gets count of closed tickets - in that context - then yes.

PR #531, opened 2013-04-03
Format root files for consistency

BafS:
I have format files from root because there is a lot of inconsistency in the code. There is still a lot to clean ;)

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -14,8 +14,8 @@
     vim: expandtab sw=4 ts=4 sts=4:
 **********************************************************************/
 
-function clientLoginPage($msg='Unauthorized') {
-    Http::response(403,'Must login: '.Format::htmlchars($msg));
+function clientLoginPage($msg = 'Unauthorized') {
+    Http::response(403, 'Must login: '.Format::htmlchars($msg));
     exit;
 }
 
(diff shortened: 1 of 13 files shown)


protich:
@BafS - Are you doing it manually or via a shell script? I'll vote for cleaning up the entire code base - now that v1.7.0 is out of the way. :+1:

BafS:
Some regexes, a bit manually and also with the eclipse formatter.
I think the code is near the 1TBS style (http://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS).
We can also follow the PSR-0, PSR-1 and PSR-2 (for style) but it would be a big re-factoring. You want it to be compatible with with PHP5+ ? (5.3+ is-it OK ?).

protich:
I'll say consistent 1TBS will do - not a big fan of some aspects of PSR. We'll work on putting together a rough doc on expected coding style going forward - I'll ping you for feedback once we have a draft.

That said, version 1.7 series supports PHP 4.4+ so PHP5 compatible styling is a no go for now. But... starting with 1.8 series, we'll require PHP 5.3+ and will include major overhaul - e.g  templating engine and ORM.

BafS:
Okay nice. Templating and ORM are a really good idea (and why not a MVC structure ?).

protich:
I'm open to MVC - templating engine will get us halfway there by separating presentation from the logic.

BafS:
Yep, a promising future.

greezybacon:
I'd like to add that 1.7 presents a URL dispatcher for the API and AJAX callbacks, so we have the _C_ part of the MVC already. So technically, ORM and template engine should get us all the way there

Issue #535, opened 2013-04-05
Internal note alerts


protich:
Internal note alerts, when enabled, are being sent out to the poster - if they happen to be ticket's last respondent or assignee.

PR #540, opened 2013-04-07
Internal note alert

protich:
Stop internal note alerts from being sent to the staff posting the note.

Addresses issue #535


greezybacon:
Looks like this is in the issue/upgrader_bugs branch already

Issue #539, opened 2013-04-07
api/task/cron vs api/tasks/cron


tomashastings:
osTicket-1.7.0/scripts/rcron.php contains the following comment:

Configuration: Enter the url and key. That is it.
url => URL to api/task/cron e.g http://yourdomain.com/support/api/task/cron
key => API's Key (see admin panel on how to generate a key)

The correct url should be http://yourdomain.com/support/api/tasks/cron (taskS, not task)

PR #541, opened 2013-04-07
Fix sample API call url path

protich:
Addresses issue #539


greezybacon:
Looks like this is in the issue/upgrader_bugs branch already

PR #564, opened 2013-05-13
Federate configuration settings

greezybacon:
Changes the config table from a column-based table to a key-value table with namespacing. This allows several parts of osTicket to store their respective configurations in the same table without requiring database structure updates.

This is going to pave the way for plugins with independent configurations as well as make things like staff- or client-based configuration easier to manage


include/class.osticket.php
```diff
@@ -64,7 +64,11 @@ function isSystemOnline() {
     }
 
     function isUpgradePending() {
-        return (defined('SCHEMA_SIGNATURE') && strcasecmp($this->getDBSignature(), SCHEMA_SIGNATURE));
+		foreach (Migrator::getUpgradeStreams() as $stream=>$hash)
```
greezybacon: This is part of the multi stream commit

PR #576, opened 2013-06-03
Rebase test suite to be more modular

greezybacon:
This makes the regression test suite more modular which will hopefully make adding more test fixtures more inviting


greezybacon:
Merged with the signal API

PR #582, opened 2013-06-13
Update class.osticket.php

tonyman19:
IMPORTANT!!!!

On some configurations - like php running as fast cgi module - both statements returns true -         

isset($_SERVER['PATH_INFO'])
isset($_SERVER['ORIG_PATH_INFO'])

but only one of them is not equal to NULL.
Please note that this "issue" will affect upgrading or installation on new systems.

I would recommend to move away from using AcceptPathInfo. To normal requests with ? Because sometimes this directive hard to turn on.

### DIFF
--- a/include/class.osticket.php
+++ b/include/class.osticket.php
@@ -337,10 +337,10 @@ class osTicket {
     }
 
     function get_path_info() {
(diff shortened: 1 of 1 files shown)


tonyman19:
The same should be applied to the master branch

PR #583, opened 2013-06-13
Fix printing / attachment download issues

greezybacon:


josh1856:
its always the little stuff confirmed fixed.

PR #585, opened 2013-06-18
Adds phone number and extension to typeahead

seagoj:
Adds an option to enable pre-population of the phone number and extension when creating a ticket.

### DIFF
--- a/include/ajax.users.php
+++ b/include/ajax.users.php
@@ -31,15 +31,31 @@ class UsersAjaxAPI extends AjaxController {
         $limit = isset($_REQUEST['limit']) ? (int) $_REQUEST['limit']:25;
         $users=array();
 
-        $sql='SELECT DISTINCT email, name '
+        $sql='SELECT DISTINCT email, name, phone, phone_ext '
             .' FROM '.TICKET_TABLE
             .' WHERE email LIKE \'%'.db_input(strtolower($_REQUEST['q']), false).'%\' '
             .' ORDER BY created '
             .' LIMIT '.$limit;
-           
+
+        global $cfg;
+
         if(($res=db_query($sql)) && db_num_rows($res)){
-            while(list($email,$name)=db_fetch_row($res)) {
-                $users[] = array('email'=>$email, 'name'=>$name, 'info'=>"$email - $name");
+
+            while(list($email,$name,$phone,$phone_ext)=db_fetch_row($res)) {
+                $record = array('email'=>$email,
+                    'name'=>$name,
+                    'info'=>"$email - $name"
+                );  
+
+                if($cfg->getSearchPhone()) {
+                    $record = array_merge($record, array(
+                        'phone' =>$phone,
+                        'phone_ext' =>$phone_ext
+                    ));
+                    $record['info'] .= " - $phone $phone_ext";
+                }
+
+                $users[] = $record;
             }                    
         }  
         
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -141,6 +141,7 @@ class OsticketConfig extends Config {
     var $defaults = array(
         'allow_pw_reset' =>     true,
         'pw_reset_window' =>    30,
+        'search_phone' =>       true
     );
 
     function OsticketConfig($section=null) {
@@ -719,6 +720,10 @@ class OsticketConfig extends Config {
         return $this->get('upload_dir');
     }
 
+    function getSearchPhone() {
+        return $this->get('search_phone');
+    }
+
     function updateSettings($vars, &$errors) {
 
         if(!$vars || $errors)
@@ -872,6 +877,7 @@ class OsticketConfig extends Config {
             'allow_email_attachments'=>isset($vars['allow_email_attachments'])?1:0,
             'allow_online_attachments'=>isset($vars['allow_online_attachments'])?1:0,
             'allow_online_attachments_onlogin'=>isset($vars['allow_online_attachments_onlogin'])?1:0,
+            'search_phone'=>isset($vars['search_phone'])?1:0
         ));
     }
 
--- a/include/staff/settings-tickets.inc.php
+++ b/include/staff/settings-tickets.inc.php
@@ -153,6 +153,14 @@ if(!($maxfileuploads=ini_get('max_file_uploads')))
             </td>
         </tr>
         <tr>
+            <td width="220">Allow Search on phone numbers</td>
+            <td>
+                <input type="checkbox" name="search_phone" <?php echo $config['search_phone']?'checked="checked"':''; ?>>
+                If enabled, client phone number can be searched and auto-filled
+                on new tickets
+            </td>
+        </tr>
+        <tr>
             <th colspan="2">
                 <em><b>Attachments</b>:  Size and max. uploads setting mainly apply to web tickets.</em>
             </th>
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -331,6 +331,10 @@ $(document).ready(function(){
             var fObj=$('#email.typeahead').closest('form');
             if(obj.name)
                 $('#name', fObj).val(obj.name);
+            if(obj.phone)
+                $('#phone', fObj).val(obj.phone);
+            if(obj.phone_ext)
+                $('#phone_ext', fObj).val(obj.phone_ext);
         },
         property: "email"
     });


seagoj:
I imagine you guys have more pressing issues, but it seems silly to not store a phone number if it was typed in. If you're not going to accept the pull request. Could you give me a quick justification?

Thanks

ntozier:
I can think of three good reasons:
1. staff who work at multiple sites and thus have multiple primary numbers numbers.
2. staff who some times work from home.
3. staff who need to specify an alternative (cell, hotel, etc) number because their primary phone is dead or they are traveling.

If users do not have to change it they wont.

Richiricheh:
I think that given the number of mods we have seen throughout the years in the mod forums there are quite a few people who'd love to have this as an option and would have it enabled by default. Personally I think this needs to be an option which can be enabled/ disabled from the admin panel. 

@ntozier I think you've brought up a interesting points and they should be considered as I have heard whispers that OStickets roadmap is potentially moving towards a profile-based UID + PWD approach (I am very much out of the loop so that might have since changed).

seagoj:
@ntozier I agree that having multiple phone numbers show up can be frustrating, but I think too much information is better than none at all. At least you have a chance of contacting the submitter if the phone number is retained.

I know everyone would use the system differently, but having the email address is nice for keeping a record of what happens, but actually being able to call is usually what gets the job done most expediently.

seagoj:
@Richiricheh I agree that it should be an option in the admin panel. I just did this quick and dirty for my own purposes. :)

In fact, that's partially why I commented on it again. If it was something people were interested in, I would happily flesh it out and make it an option (either on by default or no).

Richiricheh:
I suppose my comment was made due to the valid concerns brought forward by @ntozier. If it was a toggled option in the admin settings then I truly cannot think of a reason why it shouldn't be added as part of the project because frankly for those which want the manual phone entry then they will leave the toggle off and for those (like you and I as well as a bunch of people from the MOD forums) we would leave it toggled ON.

seagoj:
OK, this commit makes it an option in the admin settings.

It's still a bit ugly, though. I couldn't figure out the best way to access the $cfg in ajax.users.php, so it makes a separate db query. If anyone would like to point me in the right direction on how to accomplish that, I would appreciate it.

Thanks

seagoj:
Any help or is the additional call to the database the way to go?

greezybacon:
@seagoj, you should be able to:

``` php
global $cfg;
$cfg->getStorePhone();
```

seagoj:
Thanks for the help! I've edited ajax.users.php to just pull the value straight from CFG.

seagoj:
Please let me know if there's anything else that needs to be done.

setup/inc/sql/osTicket-mysql.sql
```diff
@@ -183,7 +183,8 @@ INSERT INTO `%TABLE_PREFIX%config` (`namespace`, `key`, `value`) VALUES
   ('core', 'admin_email', ''),
   ('core', 'helpdesk_title', 'osTicket Support Ticket System'),
   ('core', 'helpdesk_url', ''),
-  ('core', 'schema_signature', '');
+  ('core', 'schema_signature', ''),
+  ('core', 'store_phone', '1');
```
greezybacon: This is a bad idea. It implies a database migration. You'll also need to change `INCLUDE_DIR/upgrader/streams/core.sig` to be the MD5 hash of the updated install-mysql file. And you'll need to write a migration script for the `INCLUDE_DIR/upgrader/streams/core/` folder to migrate everyone's database to add this new setting.

We refactored the config class again. Inside the `osTicketConfig` class in `class.config.php`, there is a `$defaults` array that provides defaults for configuration not yet stored in the database. I would like to see

``` php
'store_phone' => 1
```

added to the defaults array. This method doesn't require a database migration and offers the same initial default

include/staff/settings-system.inc.php
```diff
@@ -183,6 +183,13 @@
             </td>
         </tr>
         <tr>
+            <td width="220">Retain Phone Numbers:</td>
```
greezybacon: I feel like this wording is confusing. What I read out of your code is that the phone numbers are always stored, but we're adding the capacity for the phone-number to be searchable and auto-filled on the staff-new-ticket page. Perhaps something like **Allow search on phone numbers**

greezybacon: I think this edit should go in `settings-tickets.inc.php` since it pertains to ticket management rather than system management

include/staff/settings-system.inc.php
```diff
@@ -183,6 +183,13 @@
...
+            <td>
+                <input type="radio" name="store_phone"  value="1"   <?php echo $config['store_phone']?'checked="checked"':''; ?> />Active
+                <input type="radio" name="store_phone"  value="0"   <?php echo !$config['store_phone']?'checked="checked"':''; ?> />Disabled
```
greezybacon: We commonly use a checkbox to represent the setting. I think visually it corresponds to the on/off nature of the setting better than active/disabled. It also allows space for inline help text, perhaps

```
Retain Phone Numbers:   [x] If enabled, client phone number can be searched and auto-filled on new tickets
```

include/ajax.users.php
```diff
@@ -31,15 +31,32 @@ function search() {
...
             .' FROM '.TICKET_TABLE
             .' WHERE email LIKE \'%'.db_input(strtolower($_REQUEST['q']), false).'%\' '
             .' ORDER BY created '
             .' LIMIT '.$limit;
-           
+
         if(($res=db_query($sql)) && db_num_rows($res)){
-            while(list($email,$name)=db_fetch_row($res)) {
-                $users[] = array('email'=>$email, 'name'=>$name, 'info'=>"$email - $name");
+
```
greezybacon, line 43: I think you could write this more simply like

``` php
while (list($email,$name,$phone,$phone_ext) = db_fetch_row($res)) {
    $record = array('email'=>$email, 'name'=>$name, 'info'=>"$email - $name");
    if ($cfg->getStorePhone())
        $record = array_merge($record, array(
            'phone' =>$phone,
            'phone_ext' =>$phone_ext,
        ));
    $users[] = $record;
}
```

include/ajax.users.php
```diff
@@ -31,15 +31,32 @@ function search() {
         $limit = isset($_REQUEST['limit']) ? (int) $_REQUEST['limit']:25;
         $users=array();
 
-        $sql='SELECT DISTINCT email, name '
+        $sql='SELECT DISTINCT email, name, phone, phone_ext '
```
greezybacon, line 34: This implies several hits for users with multiple phone numbers and email addresses. I think this will get cleared up when we finally get the concept of a "user" into the system. Perhaps, for now, we should instead search by unique name and then find the last email and phone number given and assume they are the most current and only show those in the search results.

include/ajax.users.php
seagoj, line 43: I've simplified this piece as you've suggested except I am appending the phone and extension to the end of $info. I've done this because I think it makes sense to retain that information as the last unique entry isn't always the one you need.

I think this makes sense in my use case, but could be turned into an search_phone_all option fairly easily.

include/ajax.users.php
seagoj, line 34: I've left this one as is because I think it's useful to have all of the unique hits displayed, but it could be made into an option, I suppose.

include/class.config.php
```diff
@@ -29,7 +29,7 @@ class Config {
     # database yet, the ->getInfo() method will not include the (default)
     # values in the returned array. $defaults allows developers to define
     # new settings and the corresponding default values.
-    var $defaults = array();                # List of default values
+    var $defaults = array('search_phone'=>1);                # List of default values
```
greezybacon: Oops. Config class is abstract and should NOT have this in the defaults list.

include/ajax.users.php
greezybacon, line 43: I agree. I missed that part in my edit. You're correct

greezybacon:
@seagoj, have you used `git rebase` before?

seagoj:
@greezybacon I have before. What would you like me to do?

greezybacon:
What we've got here is just really messy. 21 commits for a simple feature of 40 lines of code. Any way you could condense it into just a couple commits? And drop the `search_phone` default from the abstract `Config` class.

seagoj:
Ok, it's been rebased down to 1 commit and I fixed the errant search_phone in the Config class. Thanks for the help with this, by the way.

greezybacon:
I implemented this in osTicket/osTicket-1.8#2. I think this will be the best way to get it implemented.

PR #590, opened 2013-06-20
Prefer session persistent configuration over database

greezybacon:
This will allow setting something in the session and having that value override the value from the database. If persist() is called for a given key, a following get() will result in the session persisted value being returned. set() will still save the value in the database.

This is mostly useful for last-mile scenarios (and future plugins) that need to affect the configuration of the system without altering the data in the configuration table.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -49,10 +49,10 @@ class Config {
     }
 
     function get($key, $default=null) {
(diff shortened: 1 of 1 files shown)


greezybacon:
Does this imply the need for an `unpersist` method or such which will drop the key from the session?

PR #602, opened 2013-07-02
Federated email templates and site pages

protich:
Add federated email template support
Add site pages support


greezybacon:
Implemented after rebasing in #604

PR #600, opened 2013-07-03
client email not being trimmed?

greezybacon:
Multiple reports on the forums:
[Not confirmed by me personally.]

client email address with spaces at the end are not being trimmed.

report 1 text:
"It seems 3 different people were putting a space after their email address when creating a ticket. I opened the problem tickets and re-saved each email address without the space and Voila! This never happened before with version 1.6 so it's either a new bug or an incredible co-incidence.

It would be nice if the PHP code stripped blank spaces from email addresses when people submit tickets. This would certainly fix the issue."

report 2 text:
I am also facing this issue with my hMailServer + EasyPhp with osTicket-1.7.0.
But I found that users putting a space in mail id while opening new ticket.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1812,6 +1812,11 @@ class Ticket {
     function create($vars, &$errors, $origin, $autorespond=true, $alertstaff=true) {
         global $ost, $cfg, $thisclient, $_FILES;
 
+        // Drop extra whitespace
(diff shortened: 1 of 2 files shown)


greezybacon:
What a find! I can confirm that it was definitely broken

ntozier:
Vaguely related, might want to consider ensuring that the full name field disallows any response composed of all numbers.

Issue #599, opened 2013-06-30
Corrupt Attachments on Incoming emails


jonub:
As an example, an incoming email (using piping) with 2 files, one 300K .doc and an 800K pdf, both get corrupted (reduced to about half size). 

Settings: Max file size set to 5000000 bytes, max 5 files. The emails are also copied to another email address and are received fine. The problem only occurs when received by OST.

Would anyone be able to provide any sugegstions to resolve this please?

Thanks

greezybacon:
I can confirm this problem exists in the latest devel branch

greezybacon:
Any way you can test with #607?

jonub:
Great fix. Thank you very much greezybacon - see response in https://github.com/osTicket/osTicket-1.7/issues/607.

AdeliavanTonder:
Good Day

Thanks for the fix, it works great.  I was just about to give up on using OS Ticket when I found this fix last week.  I now have another issue with regards to the file attachments.  Do not know if this is related though.  When a user emails an attachment with a long file name it is not displayed as part of the attachments of the email.  All of the other attachments are displayed, just not the one with the long file name.

This is a normal .pdf file but the file name is : (23 characters).pdf Any ideas or help would be greatly appreciated.  Oh yes, when the user emails the attachment to an email program it is received just fine.

greezybacon:
@AdeliavanTonder, what was the size of the attachment? Do you have pdf extensions enabled in the admin panel?

greezybacon:
Do you have access to the raw email and would you mind sending it to me?

greezybacon:
@AdeliavanTonder, any way you could send me that raw email? Or, if you would like, we posted a fix yesterday that concerned email filename fixes (#737)  (and there's another outstanding, #738) which might address the issue. Perhaps you could retest with the tip of the develop branch.

PR #607, opened 2013-07-03
Fix attachments posted to pipe.php

greezybacon:
- Handle decoding large base64 encoded attachments better
- Avoid transcoding from one charset to the same
- Avoid encoding attachments without a declared charset


greezybacon:
Fixes issue #599

greezybacon:
I can confirm the original issue, and this seems to fix it

mohitbhatia1994:
I am having the same problem, when I download it directly from Osticket.

PR #610, opened 2013-07-05
DB Error #1064 / Template Fetch Error

greezybacon:
After updating to the latest version I am receiving the following errors when creating a new ticket via the web portal or fetching new emails to create tickets.

After using the online web ticket creation the following errors appear after submitting the ticket.
###### 

-------Start Web Form Issue------
###### 

support_issue
###### 

---Start Error Email---
###### 

db_error
###### 

---End Error Email---
###### 
###### 

---Start Email Notice---
###### 

template_error
###### 

---End Email Notice---
###### 
###### 

-------END WEB FORM ISSUE--------
###### 

There are no OUTBOUND emails sent to the third party notifying them of the issue and the ticket is actually created successfully and the system functions properly.

### DIFF
--- a/include/class.template.php
+++ b/include/class.template.php
@@ -445,8 +445,8 @@ class EmailTemplate {
         return self::lookup(self::create($vars, $errors));
     }
 
-    function lookupByName($tpl, $name, $group=null) {
-        $sql = 'SELCT id FROM '.EMAIL_TEMPLATE_TABLE
+    function lookupByName($tpl_id, $name, $group=null) {
+        $sql = 'SELECT id FROM '.EMAIL_TEMPLATE_TABLE
             .' WHERE tpl_id='.db_input($tpl_id)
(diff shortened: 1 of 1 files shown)


greezybacon:
Good find, and thanks for the detail

josh1856:
Greezybacon: When uploading a file via the web portal after applying the update I am generating the following errors.  It appears that the fetch creates no errors and the properly generates the ticket, however, the response that is supposed to be sent to the third party is not populating properly. 

Warning: preg_match_all() expects parameter 2 to be string, object given in /home/support.xyz.com/htdocs/include/class.variable.php on line 124

Fatal error: Cannot use object of type EmailTemplate as array in /home/support.xyz.com/htdocs/include/class.ticket.php on line 747

greezybacon:
I can't reopen this issue since I made a pull-request of it, I posted a fix in #611

PR #611, opened 2013-07-05
Fixup old ..MsgTemplate() getters which returned arrays

greezybacon:
This further fixes #610

### DIFF
--- a/include/class.template.php
+++ b/include/class.template.php
@@ -364,6 +364,14 @@ class EmailTemplate {
         return $this->id;
     }
 
+    function asArray() {
+        return array(
+            'id' => $this->getId(),
+            'subj' => $this->getSubject(),
+            'body' => $this->getBody(),
+        );
+    }
+
     function getSubject() {
         return $this->ht['subject'];
     }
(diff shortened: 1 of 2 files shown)


josh1856:
I pulled this and am still able to replicate the error.

########## START #########
Warning: preg_match_all() expects parameter 2 to be string, object given in /home/support.xyz.com/htdocs/include/class.variable.php on line 124

Fatal error: Cannot use object of type EmailTemplate as array in /home/support.xyz.com/htdocs/include/class.ticket.php on line 747
#########END###########

josh1856:
I was able to resolve the issue and create the ticket properly but it is still not properly being displayed in the confirmation page.

ticket_response

greezybacon:
@josh1856, I didn't merge the fix into the develop branch. Did you try it on the greezybacon:issue/610 branch or the develop?

josh1856:
greezy: I pulled the development branch and then made the changes manually.

greezybacon:
> I was able to resolve the issue and create the ticket properly but it is still not properly being displayed in the confirmation page

@josh1856, it seems to work correctly in my tests. What do you mean when you say it's not displayed properly? Are you referring to the X's?

Issue #606, opened 2013-07-03
Migrate SLA with Department and HelpTopic


greezybacon:
osTicket is currently designed so that the SLA is never changed automatically once it is set. Therefore, if the ticket changes help topics or departments, its SLA remains what was originally assigned. So even if the new department or help topic has a different SLA from the ticket being transferred, the original SLA is left intact in the ticket.

It would be nice for this functionality to be optional. I propose that for each department and for each help topic, there be an option to change the SLA of tickets to the SLA of the respective department or help topic when a ticket changed.

Therefore, when a ticket is transferred and the receiving department has this option set, the SLA of the ticket will become the SLA of the department. And likewise, when a ticket changes to a new help topic, if the option is set for the new help topic, the SLA of the ticket will be changed to the SLA of the new help topic.

akash65:
How To Do Automatically changing the SLA Plan when Help Topic or Department Changed by steps? can anyone explain me asap

PR #613, opened 2013-07-08
Allow SLA to be migrated

greezybacon:
Optionally, if enabled, the SLA of a ticket can change to the default SLA of a department or help topic when the ticket is transferred to a new department or changed to a different help topic.

This is mostly useful for users who have a generic help topic or department as the recipient of incoming tickets. Then, the SLA can be reassigned when the tickets leave the generic department or help topic

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -48,6 +48,13 @@ class Config {
         return $this->section;
     }
 
+    function getInfo() {
+        $info = array();
+        foreach ($this->config as $key=>$setting)
+            $info[$key] = $setting['value'];
+        return $info;
+    }
+
     function get($key, $default=null) {
         if (isset($this->session[$key]))
             return $this->session[$key];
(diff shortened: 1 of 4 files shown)


greezybacon:
Implements #606

greezybacon:
I wonder if this is implemented backwards. I wonder instead if a department or help topic should be flagged as 'generic'. For generic departments and help-topics, when a ticket is changed from a generic topic or department to one that isn't, the SLA should be re-evaluated. This way, if a ticket is transferred between two departments that are not generic (ie. Billing -> Shipping), the SLA will not be changed.

greezybacon:
This should be mocked up so that the SLA should be marked **Transient**. When a ticket is transferred to a new department or the help topic, the SLA will always be re-evaluated. Once the SLA is set to a non-transient one, the SLA will no longer evaluated when changing departments or help-topics.

PR #616, opened 2013-07-09
Allow generation and serving of abritrary custom pages

greezybacon:
The administrator can define pages in the admin panel defined as type `other`, and when set to active, those pages can be served from the `/pages/<page-name-slug>` URL from the base of the helpdesk.

This is mocked up only against Apache, although other servers can serve

`/pages/index.php/page-name-slug`

### DIFF
--- a/pages/index.php
+++ b/pages/index.php
@@ -1,0 +1,55 @@
+<?php
+/*********************************************************************
+    pages/index.php
+
+    Custom pages servlet
+
+    Peter Rotich <peter@osticket.com>
+    Jared Hancock <jared@osticket.com>
+    Copyright (c)  2006-2013 osTicket
+    http://www.osticket.com
+
+    Released under the GNU General Public License WITHOUT ANY WARRANTY.
+    See LICENSE.TXT for details.
+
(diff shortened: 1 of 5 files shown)


pages/index.php
```diff
@@ -0,0 +1,55 @@
...
+// Determine the requested page
+// - Strip extension
+$slug = Format::slugify($ost->get_path_info());
+
+// Get the part before the first dash
+$first_word = explode('-', $slug);
+$first_word = $first_word[0];
+
+$sql = 'SELECT id, name FROM '.PAGE_TABLE
+    .' WHERE name LIKE '.db_input("$first_word%");
```
greezybacon, line 33: This may break if the first part of the page name has non-ascii characters in it

greezybacon, line 33: This is mitigated now. `Format::slugify` will not do us-ascii transliteration (which would discard non-us-ascii characters like those from languages with a completely different alphabet). Instead, non alpha-numeric characters, regardless of the alphabet, are changed to `-`, and all letters and numbers are preserved. This method should be consistent with UTF-8 database stored in MySQL.

PR #618, opened 2013-07-10
Update settings-pages.inc.php

cdnwetzel:
on line 14 of settings-pages.inc.php the section: <a href="pages"> should read: <a href="pages.php"> link was broken


protich:
Good catch. I totally goofed on this one!

Issue #628, opened 2013-07-15
Allow login by user's email address


greezybacon:
It's a unique field for all users of the system anyway, and a large portion of users seem to be inclined to login via the email address. I propose the authentication system accept both the username and email address as identifies of the user.

PR #631, opened 2013-07-16
Allow authentication with email-address+password

greezybacon:
Previously, username and password were required. With this patch, the email address can be used to authenticate the user with the password


greezybacon:
Implements #628

PR #632, opened 2013-07-16
Allow custom logo integration on client site

greezybacon:
This patch allows a user to upload an image to be displayed in the header of the client interface. The logo is stored in the database as a file with ft='L'. This also changes the `AttachmentFile::deleteOrphans` function to only consider aTtachment file types where `ft=='T'`

It introduces a `logo.php` and `scp/image.php` script to serve file content from the database, and `AttachmentFile::display` now supports client-side caching with the `Last-Modified`, `ETag`, `If-Modified-Since`, and `If-None-Match` headers. `AttachmentFile::display` will return HTTP/304 if the client-side cache should be used and doesn't download the data from the database.


include/class.config.php
```diff
@@ -863,13 +877,32 @@ function updatePagesSettings($vars, &$errors) {
...
+                $errors['logo'] = 'Unable to upload logo image. '.$error;
+        }
+
         if(!Validator::process($f, $vars, $errors) || $errors)
             return false;
 
+        if (isset($vars['delete-logo']))
+            foreach ($vars['delete-logo'] as $id)
+                if (($f = AttachmentFile::lookup($id)))
+                    $f->delete();
```
greezybacon, line 898: Ensure logo is not selected before deleting

include/class.file.php
```diff
@@ -172,15 +196,46 @@ function upload($file) {
         return AttachmentFile::save($info);
     }
 
+    function uploadLogo($file, &$error, $aspect_ratio=3) {
+        /* Borrowed in part from
+         * http://salman-w.blogspot.com/2009/04/crop-to-fit-image-using-aspphp.html
+         */
+        $source_path = $file['tmp_name'];
+
+        list($source_width, $source_height, $source_type) = getimagesize($source_path);
```
greezybacon, line 209: Check for **gd** extension

PR #638, opened 2013-07-21
Password reset feature

greezybacon:
Uses a seven step procedure:
1. (user) Fails to login twice or more
2. Clicks the 'Forgot my password' link on the login form
3. Submits the username or email address and triggers a password-reset
    email
4. Clicks the link in the email and is directed back to the reset page
5. Enters the username or email again and is logged in
6. Password change is forced, but current password is not required
7. Password is updated, user can continue the session without
    authenticating again

**Outstanding**
- ~~SQL edit and migration script to add a default template for the password reset email~~

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -15,6 +15,7 @@
 **********************************************************************/
 include_once(INCLUDE_DIR.'class.ticket.php');
 include_once(INCLUDE_DIR.'class.dept.php');
+include_once(INCLUDE_DIR.'class.error.php');
 include_once(INCLUDE_DIR.'class.team.php');
 include_once(INCLUDE_DIR.'class.group.php');
 include_once(INCLUDE_DIR.'class.passwd.php');
@@ -401,6 +402,7 @@ class Staff {
 
     //Staff profile update...unfortunately we have to separate it from admin update to avoid potential issues
     function updateProfile($vars, &$errors) {
+        global $cfg;
 
         $vars['firstname']=Format::striptags($vars['firstname']);
         $vars['lastname']=Format::striptags($vars['lastname']);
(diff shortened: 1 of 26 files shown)


include/staff/settings-system.inc.php
```diff
@@ -126,10 +131,20 @@
...
+              <input type="checkbox" name="allow_pw_reset" <?php echo $config['allow_pw_reset']?'checked="checked"':''; ?>>
+              <em>Enables the <u>Forgot my password</u> link on the staff
+              control panel</em>
+            </td>
+        </tr>
+        <tr><td>Password Reset Window:</th>
+            <td>
+              <input type="text" name="pw_reset_window" size="6" value="<?php
+                    echo $config['pw_reset_window']; ?>">
+                Maximum time between the password reset email and when the
```
greezybacon: Specify the time is in **minutes**

include/class.staff.php
```diff
@@ -660,6 +683,43 @@ function create($vars, &$errors) {
...
+        $msg = $ost->replaceTemplateVariables($template->asArray(), array(
+            'url' => $ost->getConfig()->getBaseUrl(),
+            'token' => $token,
+            'reset_link' => sprintf(
+                "%s/scp/pwreset.php?token=%s",
+                $ost->getConfig()->getBaseUrl(),
+                $token),
+        ));
+
+        if(!($email=$cfg->getAlertEmail()))
```
greezybacon, line 714: Shouldn't this be inside some magic function like

``` php
$ost->sendAlert($this->getEmail(), $msg['subj'], $msg['body']);
```

or even

``` php
$ost->sendAlert($this->getEmail(), $template, array('url'=> ...));
```

PR #652, opened 2013-07-30
Properly escape HTML chars in text/plain email bodies

greezybacon:
If there characters in the plain text body of the email that appear like
HTML tags, for instance

```
From: <sip:527772432@172.18.0.2>;tag=952422a9dd1ap1a6o1
```

The <sip...> part would be removed by the Format::striptags() call in
Format::sanitize().


greezybacon:
**TODO** There is a similar issue concerning popup for the "System Logs" page where the text in the popup is not properly escaped

PR #654, opened 2013-08-01
Make default data translatable

greezybacon:
This is one of the last major hurdles to make osTicket fully translatable.
With the advent of 1.7.0, database migration is based on the MD5 hash of the
install SQL file. Therefore, the install SQL script cannot be modified (ie.
translated) or else the system cannot be migrated, since the MD5 hash will
no longer match up.

This patch pulls initial data out of the install file so that it can be
translatable. Eventually the install worksheet can allow a drop-down box of
the default system language. The default data from that language will be
loaded into the system.


greezybacon:
We need to discuss breaking out the default configuration
- Which parts go in the install sql
- Which parts should be translatable
- Which parts should be defaulted in the code (changes don't break migration but don't need to be translated)

protich:
Branch deferred to 1.8 repo.

PR #688, opened 2013-09-02
Remove requirement of external-ticket-id in subject line

greezybacon:
This patch affords an administrator the ability to remove the [#%{ticket.number}] from the email template subject line for the new ticket autoresponse and the new message autoresponse. Previously, the ticket number with a prefixed hash in brackets was used to identify which ticket thread an email was in reference to.

With this patch, the email message-id (which was already kept on file) is sent in the MIME "References" header. When a user responds to an autoresponse email, the "References" will include this message-id in the return email. The ticket thread is then matched up with the email based on the message-id rather than the subject line.

Ticket numbers are still supported in the subject line, in the event that non-compliant email clients do not properly include the References header.

### DIFF
--- a/include/api.tickets.php
+++ b/include/api.tickets.php
@@ -113,6 +113,10 @@ class TicketApiController extends ApiController {
                 return $ticket;
         }
 
+        if (($thread = ThreadEntry::lookupByEmailHeaders($data))
+                && $thread->postEmail($data)) {
+            return true;
+        }
         return $this->createTicket($data);
     }
 
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -423,11 +423,6 @@ class ApiEmailDataParser extends EmailDataParser {
         if(!$data['emailId'])
             $data['emailId'] = $cfg->getDefaultEmailId();
 
-        if($data['email'] && preg_match ('[[#][0-9]{1,10}]', $data['subject'], $matches)) {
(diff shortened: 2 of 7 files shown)


PawelBaczkiewicz:
Hi,

In new announcement of osticket 1.7.2 there was some part saying:
"However, matching on the external ticket number is still supported as a fallback in the event you have some really old, non-compliant client email software in use. "

I have to say, that it seems to not to work.
I have some answers to my ticket connected with my ticket correctly, but sometimes (about 20% of all tickets) there is new ticket created with topic having old ticket number.
Those "wrongly created" tickets were created in new version.

For example. Today osTicket has created new ticket with topic:

Ticket id: 752352 , topic: Email problems

I answered to this ticket, and customer answered immediatelly and new ticket was created:

Ticket id: 532168, topic: Re: [BOK#752352]: Email problems

So probably his software does not support this new thing with hidden ID in header of email. But as I understood this internal ticket ID should be still supported.
Should I do something with configuration ? I did not change anything just upload new version (upgrade from 1.7.1 to 1.7.2).

Wrongly created new tickets instead of connecting answers to old ticket using internal ticket id (if there is no hidden value of ticket in email header) have also impact on topics.
Good mechanism of osticket is adding internal ticket id using pattern defined in administration mode.
But if this mechanism works not correctly you can have some bugs like this:

RE: [BOK#778804]: Re: [BOK#357566]: Re: [BOK#183743]: Something

Above line is from my osticket - of course - customers see this in their email topics

PR #692, opened 2013-09-03
Update client.inc.php incorrect behaviour in offline mode

sabas:
With ./, if called from knowledge base, it searches incorrectly offline.php in that folder.


sabas:
Okay, @greezybacon I tried with ROOT_DIR, I noted that I was trying with ROOT_PATH (which caused error), committed the fix :)

PR #696, opened 2013-09-03
Don't send emails for non-existent table

greezybacon:
When an admin logs in to upgrade to 1.7.1 and further from a version pervious
to 1.7.1, the system will attempt to clear password reset tokens from the
config table, which hasn't been upgraded yet to the namespaced version from
1.7.1

This patch avoids sending an email with the offending SQL statement to the
administrator.


greezybacon:
I have to wonder if we should just disable db error emails when the upgrade is pending or in progress

Issue #672, opened 2013-08-20
Staff Login


ghost:
Installed the new os Ticket 1.7.1 when logging in control panel it gives an error valid csrf required. This is a fresh install. Any solutions

sabas:
@greezybacon same problem as mine in #653 , I can reproduce in a plain installation.

greezybacon:
@merali780, @sabas, would either of you be available and willing to do a screencast session with me -- perhaps with TeamViewer or GoToMeeting? I haven't been able to reproduce the CSRF token issue.

sabas:
@greezybacon also #677. Sorry I can't do screencast here at work, but I could try if I replicate the error at home. (Or if you want to try directly something drop me an email or an Hangouts chat, I have them open.

greezybacon:
@merali780, are you accessing the osTicket installation on localhost by chance?

pedronoliveira:
Having the exact same error on a localhost test instalation, tested with localhost address and throught other workstation on the same network as my localhost throught private IP, still get the same error.

EDIT: Just tested the same setup on my private VPS ArchLinux Lighttpd Mysql and it works normally

jimmyeao:
im having this issue on a site I have tried to upgrade from 1.6 - running on Apache2.2.22 under debian in a VM. PHP Version 5.4.4-14+deb7u2 MySQL client version: 5.5.31
So far have tried to access via localhost, private ip, DNS name
Have downloaded the latest 1.7 dev release, still no joy..

greezybacon:
@jimmyeao there was another issue concerning the detection of the URL path for the osTicket install -- #716. It was merged today. Would you mind trying again? We pushed the version out to the osticket.com site, so you could re-download it if you would like.

jimmyeao:
Still have the same error... I installed the Dev version and the version of the osticket site to be sure...
I did notice a new requirement to install php-pear when installed the dev release

jimmyeao:
Happy to do a remote session tomorrow around this time if that would help?

greezybacon:
@jimmyeao, sorry about the php-pear issue. It was a mistake on our part. It is definitely not required. We've addressed that this week in the develop branch and in a new release

greezybacon:
Can anyone / everyone verify that this is fixed with the tip of the develop branch and in osTicket 1.7.1.4?

andresadanmx:
[...]

Issue #653, opened 2013-07-31
Persistent session not working? a possible solution


sabas:
Every time I accessed the admin panel (/scp/), I had to relogin again, so here it is a possible solution

I found that moving line 38 in main.inc.php
   //ini_set('session.cookie_path','/osticket/');
to line 74 after 
   define('SQL_DIR', UPGRADE_DIR.'sql/');

and rewriting it as 
ini_set('session.cookie_path',ROOT_PATH);

recovers the session and no more login panel every time.

I don't send a pull request because I don't know why that line was commented..

greezybacon:
Thanks for identifying this. We've actually had a considerable amount of trouble out of the bug. I have mocked up the issue as follows:

``` php
$is_https = (isset($_SERVER['HTTPS'])
        && strtolower($_SERVER['HTTPS']) == 'on')
    || (isset($_SERVER['HTTP_X_FORWARDED_PROTO'])
        && strtolower($_SERVER['HTTP_X_FORWARDED_PROTO']) == 'https');

session_set_cookie_params(86400, dirname($_SERVER['PHP_SELF']),
    $_SERVER['HTTP_HOST'], $is_https);
```

Could you retest with something like the above code?

greezybacon:
What browser did you observe this in by chance?

sabas:
Latest Chrome!
I'll test tomorrow morning, because I have the installation in the pc I'm using at work.. That one is a xampp distribution on windows xp (now I'm on linux, I can try also here)

Edit: as a matter of fact, chrome on linux I don't notice the bug.. :/ The patch (put on line 38) caused not to login with error "session time out due to inactivity"

sabas:
With the solution https://github.com/osTicket/osTicket-1.7/issues/653#issuecomment-21962384 I couldn't login ("Authentication required"). The pull request same (btw the code from that branch seems different from the develop branch' 1.).
I also tried to delete and reinstall (I'm using 1.7 stable + the bootstrap theme from the forums)

greezybacon:
@sabas, did you test with the branch I posted in #655?

> (btw the code from that branch seems different from the develop branch' 1.).

I'm not following quite you mean

sabas:
@greezybacon yep I tried that, sorry I badly wrote the message..
I tried both patching following #655 pull request and reinstalling directly that branch. I didn't see cookies also (don't know if it's an useful info, but the way I tried produces in Chrome a PHPSESSID cookie with the correct folder)

[...]

PR #709, opened 2013-09-06
Fix cookie domain for localhost

### LINKED ISSUE #677: Valid CSRF Token Required
Hi,

I recently upgraded mine from 1.7 to 1.7.1 then I got "Valid CSRF Token Required", every time I login.

greezybacon:
Web browsers don't appreciate a cookie domain without any dots. This patch
detects the originally-requested domain for the request. If the domain does
not contain dots (such as 'localhost' or the name of a local server on your
network defined in your hosts file), no cookie domain is sent.

The greatest symptom of this issue what the illustrious 'Invalid CSRF token'
seen repeatedly on the scp login page. The reason is that the browser was
rejecting the cookie from the server.

Fixes #677, #672, #653

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -25,19 +25,32 @@ class osTicketSession {
         if(!$this->ttl)
             $this->ttl=SESSION_TTL;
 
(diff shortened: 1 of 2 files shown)


sabas:
Whohoo!
3ef_woohoo

PR #710, opened 2013-09-06
Fix for "Warning: strpos(): Empty needle"

akcoder:
Fix a minor php warning (comes about because we run our dev server in strict mode with all warnings turned on)

### DIFF
--- a/include/class.misc.php
+++ b/include/class.misc.php
@@ -149,12 +149,16 @@ class Misc {
         $path = '';
         while (strpos($_SERVER['DOCUMENT_ROOT'], $root) === false) {
             $lastslash = strrpos($root, '/');
-            if ($lastslash === false)
+            if ($lastslash === false) {
                 // Unable to find any commonality between $root and
                 // DOCUMENT_ROOT
                 return './';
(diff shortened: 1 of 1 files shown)


greezybacon:
Would you mind submitting the values of `$_SERVER['DOCUMENT_ROOT']` and the value of `$main_inc_path` from the top of the `siteRootPath` function?

akcoder:
doc root: /srv/httpd/htdocs'
$main_inc_path: /home/dmorphis/public_html/osTicket-1.7

greezybacon:
@akcoder why did you close the request?

akcoder:
Because there was some things with my fix that didn't work right. For
example, on the landing page if you clicked on any link repeatedly the url
would end up as http://somesite/someurl/osTicket//////open.php. I fixed
that locally by returning '.', but I'd like to take some time to explore a
better solution.

Plus I don't like how github does pull requests. I have another one coming
with a completely different feature, and github would put it into the
existing pull request. Maybe there is a way around that, I just haven't
taken the time to explore it yet.

-dan

PR #711, opened 2013-09-06
Allow HTTP auth pass-though

akcoder:
Allow HTTP auth pass-through. Useful where you have a paranoid systems admin who only trusts HTTP auth to protect the site. Removes the logout link if the user authenticated with HTTP auth.

Also fixes what I presume was a minor oversight where "Howdy, username" was displayed instead of the users full name.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -719,6 +719,62 @@ class OsticketConfig extends Config {
         return $this->get('upload_dir');
     }
 
+    /**
+     * Returns true if HTTP pass-though authentication is enabled, otherwise false.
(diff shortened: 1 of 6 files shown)


protich:
Salutation fixed in pull request #713. Opted to show just the first name, full name on salutation might break the page layout for users with long names.

We will revisit authentication is 1.8 or 1.9 - the plan is to provide pluggable authentication plugin system with authentication backend such as HTTP auth, LDAP etc.

akcoder:
Thank you for the update!

PR #713, opened 2013-09-09
Remove "Howdy" as the salutation - no sense of humor in this world!

protich:


include/staff/header.inc.php
```diff
@@ -41,7 +41,7 @@
     ?>
     <div id="header">
         <a href="index.php" id="logo">osTicket - Customer Support System</a>
-        <p id="info">Howdy, <strong><?php echo $thisstaff->getUserName(); ?></strong>
+        <p id="info">Welcome, <strong><?php echo $thisstaff->getFirstName(); ?></strong>
```
greezybacon, line 44: Why not

``` html+php
Yo dawg, waddup <?php echo $thisstaff->getLastName(); ?>?
```

military guys would appreciate that

PR #717, opened 2013-09-11
Turn off display_errors for releases

greezybacon:
Ensures that the settings for `display_errors` and `display_startup_errors` are disabled (if specified) for releases

### DIFF
--- a/setup/cli/package.php
+++ b/setup/cli/package.php
@@ -132,7 +132,11 @@ chdir($stage_path);
 
 // Replace THIS_VERSION in the stage/ folder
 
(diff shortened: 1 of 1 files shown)


ntozier:
display_errors have been off longer than I have been using osTicket.  It
would sure be nice if that setting was moved out of main.inc.php (lines
49-50) and into admin panel though. :)  It would be even better if you
could check a box and only have display errors stay on for X amount of time
(say 10 minutes)

greezybacon:
I agree, Neil. We're redesigning the configuration in osTicket 1.8 which will make everything the administrator would like to change possible, without adding 100 more options to the web interface. Another example would be the minimum time between autocron runs (3 minutes currently). There are many other nit-picky settings that should be changeable, but don't necessary need to exist in the web interface.

PR #765, opened 2013-10-06
Better converting from ticket thread to PDF

### LINKED ISSUE #756: HTML codes in Print (pdf)
When printing ticket from 1.7.1.4 the Danish letters æøå, > and a few more is shown as html codes in the PDF.

greezybacon:
Assume that text in the database is encoded in UTF-8 and assume that it is HTML text and the entities should be decoded prior to display in the PDF.

Fixes #756

### DIFF
--- a/include/class.pdf.php
+++ b/include/class.pdf.php
@@ -91,11 +91,21 @@ class Ticket2PDF extends FPDF
     }
 
     function _utf8($text) {
+        // Assume text is in utf-8 charset
+        $flags = ENT_COMPAT;
+        if (phpversion() >= '5.4.0')
+            $flags |= ENT_HTML401;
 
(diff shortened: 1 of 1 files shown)


OleSta:
Danish letters works with iconv but not without (uncomment line 102 and 103 to test).
fpdf does not work with utf8.

greezybacon:
I agree. I used utf8_decode as a fallback option (which still may not be available)

OleSta:
Is this correct?
$flags |= ENT_HTML401; 
Should it be?:
$flags =| ENT_HTML401; 
or
$flags = $flags | ENT_HTML401;

greezybacon:
`|=` is correct. It is a short form of

``` php
$a = $a | $b;
$a |= $b;
```

PR #767, opened 2013-10-09
Mark tickets overdue which have no SLA

### LINKED ISSUE #757: Overdue not flagged with cron on duedate
I found that in OSTicket 1.7.1 (downloaded on 10/2/2013) the overdue flag and notification where not sent on tickets where the duedate had passed.

When I looked in the checkOverride method of class.ticket.php I found that the SQL query seemed to be attempting to pull a list of tickets to flag overdue and one of the options was if the ticket's duedate had passed. In this case no SLA was set, so the inner join in the query was not allowing the ticket to return from the query as no SLA was assigned.

The Fix:

I changed the join in that query to a left join rather than an inner join and the tickets with duedates in the past that have not already been marked overdue and had no assigned SLA are processed as expected.

greezybacon:
If a ticket has a due date and does not have an SLA set, the ticket would never transition to overdue. This patch addresses an SQL flaw preventing the ticket from being transitioned.

Fixes #757

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2166,7 +2166,7 @@ class Ticket {
     function checkOverdue() {
 
         $sql='SELECT ticket_id FROM '.TICKET_TABLE.' T1 '
-            .' INNER JOIN '.SLA_TABLE.' T2 ON (T1.sla_id=T2.id AND T2.isactive=1) '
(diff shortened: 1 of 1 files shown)


mcotterman:
Looks right to me.

Thank you!

PR #769, opened 2013-10-10
Adding support for database connection on a non-standard port.

ghost:
Just thought it would be of some use.


greezybacon:
@infoza, thanks for your recommendation here. I don't mean to preach, but I just couldn't consider accepting your request as you wrote it because you practically rewrote the `mysqli.php` file, including the header. I've made a separate pull request to implement the feature in #775

PR #802, opened 2013-10-25
Included LDAP Config along with Pending Ticket Status

jdelhome3578:


greezybacon:
A single pull request should propose a single, coherent, complete idea

PR #804, opened 2013-10-25
Pending Ticket Status along with HTML email template

jdelhome3578:
needs updated to new syntax


greezybacon:
HTML email templates are already implemented in osTicket/osTicket-1.8#5

PR #806, opened 2013-10-25
Order by department

jdelhome3578:


greezybacon:
I agree with this change; however, you need to rebase it against the tip of the develop branch before I could accept it

PR #807, opened 2013-10-25
HTML email template

jdelhome3578:


greezybacon:
This code was moved to `class.mailer.php`

PR #808, opened 2013-10-25
Included LDAP Config

jdelhome3578:


greezybacon:
This will be implemented in osTicket-1.8 along with pluggable authentication which will allow for authentications other than built-in and LDAP

PR #811, opened 2013-10-25
Included LDAP Config

jdelhome3578:


greezybacon:
You need to stage your code changes that go together, like "LDAP Config" into one pull request. Again, this will be implemented in osTicket-1.8

PR #820, opened 2013-10-25
Create class.ldap.php

jdelhome3578:


greezybacon:
You need to stage your code changes that go together, like "LDAP Config" into one pull request

PR #824, opened 2013-10-28
image path fix - Update scp.css

ntozier:
fix broken background image which appears was moved from
../images/kb_category_bg.png
to
../assets/default/images/kb_category_bg.png


greezybacon:
This is fixed with #831

PR #848, opened 2013-11-15
Pending Ticket Status

jdelhome3578:


clonemeagain:
Somehow I doubt this will get pulled, it changes waay too much in one go.. And reverts major parts of the config..  Is it tested?

jdelhome3578:
This setup is in production and was built under 1.7.0

clonemeagain:
But the pull would be against 1.7.5 which uses a different syntax for config etc.. hence the vast number of changes..

See https://github.com/osTicket/osTicket-1.7/pull/848/files

jdelhome3578:
If somebody can port the changes to the new version and also 1.8 that would
be great.

PR #849, opened 2013-11-17
API JSON attachments

fanha99:
this fix for the API JSON attachments which is match the "JSON Payload Example" with RFC 2397.
which is in the document here:
https://github.com/osTicket/osTicket-1.7/blob/develop/setup/doc/api/tickets.md

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -363,14 +363,14 @@ class ApiJsonDataParser extends JsonDataParser {
             } else if ($key == "autorespond") {
                 $value = (bool)$value;
             } else if ($key == "attachments") {
-                foreach ($value as &$info) {
-                    $data = reset($info);
+                foreach ($value as $name=>&$info) {
+                    $data = $info;
                     # PHP5: fopen("data://$data[5:]");
                     if (substr($data, 0, 5) != "data:") {
                         $info = array(
                             "data" => $data,
                             "type" => "text/plain",
-                            "name" => key($info));
+                            "name" => $name);
                     } else {
                         $data = substr($data,5);
                         list($meta, $contents) = explode(",", $data);
@@ -378,7 +378,7 @@ class ApiJsonDataParser extends JsonDataParser {
                         $info = array(
                             "data" => $contents,
                             "type" => ($type) ? $type : "text/plain",
-                            "name" => key($info));
+                            "name" => $name);
                         # XXX: Handle decoding here??
                         if (substr($extra, -6) == "base64")
                             $info["encoding"] = "base64";


fanha99:
with that document, still confuse about the document and the example in it:

the document in the top of page:

attachments: An array of files to attach to the initial message. Each attachment must have some content and also the following fields:

```
name: required name of the file to be attached. Multiple files with the "same name"  are allowable
type: Mime type of the file. Default is text/plain
encoding: Set to base64 if content is base64 encoded
```

but the example is:

```
    "attachments": [
        {"file.txt": "data:text/plain;charset=utf-8,content"},
        {"image.png": "data:image/png;base64,R0lGODdhMAA..."},
    ]
```

where is the fields: name, type, encoding?
with the example like this: the file name as key, how can I up load 2 files with the same name?

greezybacon:
Filetype and encoding are placed in the rfc 2397 encoding:

```
data:text/plain;charset=us-ascii;base64,...
```

Where "text/plain" is the file type, charset declares a charset (ISO-8859-1 is assumed otherwise for text/\* types), and base64 indicates that the body is encoded using base64. The ... following the comma represents the body of the attachment (base64 encoded in this example).

The attachments is a list of JSON objects, which is distinct from a single JSON object so the the key of the object could be repeated. The key is the file name.

This format may be confusing, but I feel that it embodies the simplicity of JSON formatting. If you want something more structured, use the XML payload.

If you want to mock up a change to the documentation to make it clearer, I would be willing to discuss pulling it

Cheers,

Von meinem iPhone gesendet

fanha99:
hi, thx for explain.

I built the attachments like this:

```
$attachments[$file['name']] = 'data:' . $file['type'] . ';base64,' . base64_encode(file_get_contents($file['tmp_name']));
```

that why could not match the API.

it must be:

```
$attachments[] = array ($file['name'] => 'data:' . $file['type'] . ';base64,' . base64_encode(file_get_contents($file['tmp_name'])));
```

so, please close this request.

BTW, when I read this block of text

```
 attachments: An array of files to attach to the initial message. Each attachment must have some content and also the following fields:

    name: required name of the file to be attached. Multiple files with the same name are allowable
    type: Mime type of the file. Default is text/plain
    encoding: Set to base64 if content is base64 encoded
```

I was thinking about building the attachment like this:

```
                    // $attachments[] = array (
                        // 'name' => $file['name'],
                        // 'type' => $file['type'],
                        // 'data' => base64_encode(file_get_contents($file['tmp_name'])),
                        // 'encoding' => 'base64',
                    // );
```

PR #854, opened 2013-11-20
bug fix for custom upgrade streams

alpianon:
line 199: in case of custom upgrade streams, variable $signature contains the last value from the foreach cycle at line 130, not the core schema signature!

line 209: if we do not insert schema signatures for each custom stream, after install osTicket thinks it needs an upgrade and gets stuck (see function isUpgradePending() in class.ticket)


alpianon:
I found this bug because I tried to create a custom stream for my osticket mod
With this patch it works
You can reproduce the bug by creating a dummy custom stream and including it in streams.cfg, and then trying to install osticket. Installation will seem ok, but when you try to login to /scp you get an error message that an upgrade is pending

greezybacon:
After reviewing it further, I feel that your implementation written as nicely as possible for 1.7. I'll deal with the merge conflict when porting it to 1.8 and implement it with the Config class as shown in my previous reply.

Thanks for posting this. Seriously.

PR #856, opened 2013-11-22
Faster SQL-query

eivindgit:
Faster SQL-query, huge performance improvement on installations with thousands of tickets. We use osTicket and have about 25 000 tickets and it was getting slow.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1797,46 +1797,53 @@ class Ticket {
         if(!$staff || (!is_object($staff) && !($staff=Staff::lookup($staff))) || !$staff->isStaff())
             return null;
 
-        $sql='SELECT count(open.ticket_id) as open, count(answered.ticket_id) as answered '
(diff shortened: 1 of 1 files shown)


greezybacon:
This looks good. Thanks for taking the time to propose it

eivindgit:
Thank you for bringing life into osTicket again! We were looking into changing system before the development startet up again!

greezybacon:
Absolutely no offense, @eivindgit, I confirm that your query runs about ~380ms less (~390ms down to ~10ms) on a system with about 9k tickets, or about 40x faster. This is huge. Thanks so much.

Where clauses with '1 OR ...' or '1=1 OR ...' give me hives and just look like SQL injection. Also, your proposal might use `$where2` uninitialized. Instead of bugging you about it, I pulled your commit, made the changes and proposed #862.

eivindgit:
You are totally right @greezybacon, thanks for the code review!

PR #866, opened 2013-12-13
Fixes FAQ time issue 136

clonemeagain:
Need to run these queries to update the schema, should perhaps be your call how that is done:

```
ALTER TABLE `ost_faq` CHANGE `created` `created` DATETIME NOT NULL 
ALTER TABLE `ost_faq` CHANGE `updated` `updated` DATETIME NOT NULL 
```

Original was trying to display the last update of the category, instead of the faq item. Tested, works.

Refer: https://github.com/osTicket/osTicket-1.8/issues/136

### DIFF
--- a/include/client/faq.inc.php
+++ b/include/client/faq.inc.php
@@ -29,4 +29,4 @@ if($faq->getNumAttachments()) { ?>
 </div>
 </p>
 <hr>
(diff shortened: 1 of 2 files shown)


greezybacon:
Thanks for you investigative work here. Would you mind rebasing the two commits to use `Format::db_date`? I will add a patch to be released in 1.8.1 to change the database columns

PR #872, opened 2013-12-23
Fixes XXXX ticketid error

clonemeagain:
Forum post: http://www.osticket.com/forum/discussion/74564/ticket-numbers-showing-xxxxxxx/p1

Still a problem, but now its due to the template itself having a variable that it shouldn't.

Suggest SQL patch of same for 1.7.5 upgrade to remove message from web-created-tickets.

Could possibly create admin option to display extId, but would have to modify /open.php and remove the ${ticket.extId} from the array and change the fill to 2.. tested, works.  This seems by design, so reworked my solution to simply remove the errant variable.

Workaround for already installed versions: As admin, visit: /scp/pages.php?id=2

Change the text to read:

```
%{ticket.name},
Thank you for contacting us.

A support ticket has been created and a representative will be getting back to you shortly if necessary.

Please check your email for login information.

Support Team
```

### DIFF
--- a/setup/inc/streams/core/install-mysql.sql
+++ b/setup/inc/streams/core/install-mysql.sql
@@ -1,4 +1,3 @@
-
 DROP TABLE IF EXISTS `%TABLE_PREFIX%api_key`;
 CREATE TABLE `%TABLE_PREFIX%api_key` (
   `id` int(10) unsigned NOT NULL auto_increment,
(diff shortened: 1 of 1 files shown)


greezybacon:
The install script isn't so easily edited. If you want to migrate 1.7 users to 1.8 and drop the confusing text and tokens, then the script should go in the 1.8 repo under `include/upgrader/streams/core/16fcef4a-d51f303a.patch.sql`

clonemeagain:
Problem doesn't exist in 1.8, but with a  impending 1.7.5 it seemed prudent to get this in there. Might have to wait till after Christmas.

PR #878, opened 2014-01-17
Fix parameter count to TicketLock::lookup

greezybacon:
This is basically a slight rewrite of @clonemeagain's patch to remove one of the lines of changes


clonemeagain:
I feel bad for not refactoring this myself. Still learning. Next time!

